PR #123, opened 2011-12-09
README: explain cron syntax and time zones.

henrik:
As I mentioned on Twitter, resque-scheduler/rufus-scheduler interprets the cron syntax as server-local time instead of Rails' config.time_zone. This commit updates the README to explain that gotcha with a workaround.

Hopefully I, or someone else, will find the time to make this happen automatically, but until then, this might be helpful.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -189,6 +189,24 @@ A big shout out to [rufus-scheduler](http://github.com/jmettraux/rufus-scheduler
 for handling the heavy lifting of the actual scheduling engine.
 
 
+#### Time zones
+
+Note that if you use the cron syntax, this will be interpreted as in the server time zone
+rather than the `config.time_zone` specified in Rails.
+
+You can explicitly specify the time zone that rufus-scheduler will use:
+
+    cron: "30 6 * * 1 Europe/Stockholm"
+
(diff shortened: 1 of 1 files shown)


shawnstephens:
This section is no longer accurate. As of Rufus Scheduler 3.3.3...

> Rufus-Scheduler 3.3.3 detects the presence of Rails and uses its timezone setting (tested with Rails 4), so setting ENV['TZ'] should not be necessary.

diei:
It is correct what @shawnstephens wrote. Since rufus-scheduler 3.3.3 it determines the timezone from Rails default with the gem tzinfo, if Rails is present. If Rails is not present the timezone is determined from system.

But since rufus-scheduler 3.4.0 the mechanism has changed again. The determination of the timezone was extracted to the gem et-orbi (using also tzinfo) and since that the fallback getting the timezone from system was removed.

The README should be updated with the info, that the default timezone is determined from the Rails config. If you have no Rails the timezone is tried to be determined by `EtOrbi.determine_local_tzone`, but that does not work correctly (in my oppinion) and any timezone could be the result, except you have set `ENV['TZ']`.

PR #463, opened 2015-01-07
Check alive in the safe way

dtaniwaki:
Hi,

It seems that `@th` is not defined yet if the process gets the signal as soon as it starts. I think we don't want to get an exception for this.

[Review on Reviewable](https://reviewable.io/reviews/resque/resque-scheduler/463)


yaauie:
Thanks!

Issue #468, opened 2015-03-05
Delayed jobs are not being run


danlo:
Greetings,

I am running resque-scheduler (4.0.0), resque (1.25.2), ruby (2.2.0p0)

I run the resque-scheduler standalone job via

LOGFILE=/[censored]/resque-scheduler.log VERBOSE=true bundle exec resque-scheduler -B

And, my Resque.enqueue_at() jobs are not being executed when they should be:

![capture](https://cloud.githubusercontent.com/assets/525181/6510090/9d27dc2e-c318-11e4-930d-47a2ed5b266a.JPG)

Do you have any thoughts on how to go about debugging why my jobs are not being executed.

I've tried restarting resque-scheduler several times, but it hasn't executed the new jobs.  I have no "cron" like scheduled jobs.

(The first block of jobs, were removed and manually executed)

-daniel

yaauie:
It looks like the "Job count" column is 0 for times in the past; for some reason that isn't getting cleaned up, but the jobs are getting removed from the delayed queue; looking at the code, I see that error handling in enqueueing a job happens [*after*](https://github.com/resque/resque-scheduler/blob/v4.0.0/lib/resque/scheduler.rb#L195-L199) the job is popped from the delayed queue, and that exceptions raised here are squashed & logged. Do your logs indicate that anything is amiss?

The command-line utility `less` can be used to inspect your log file; the '+G' flag will cause it to start showing the end-of-file, which may be where the interesting stuff is happening, but you can scroll with the arrow keys:
``` sh
$ less +G /[censored]/resque-scheduler.log
```
It's possible that your scheduler process doesn't have access to the job's class, which would cause it to raise an exception, or that something in the job class' `before_enqueue_*` hook is raising an exception; the logs will tell you more.

Your command above doesn't include [`--initializer-path`](https://github.com/resque/resque-scheduler/blob/v4.0.0/lib/resque/scheduler/cli.rb#L48-L50) or [`--environment`](https://github.com/resque/resque-scheduler/blob/v4.0.0/lib/resque/scheduler/cli.rb#L44-L45) flags, so I assume the process just doesn't know enough about the work to be able to put it on a queue (it needs the class definition for that). If this is the case, you'll need to point it to a ruby script that can load your application (without starting any server processes), at least far enough to have all the relevant job classes loaded.

Can you verify that the jobs are getting placed on their queues?

danlo:
Ah!  That was was the problem

There wasn't an initializer-path.

[...]

PR #469, opened 2015-03-05
Note about Resque.redis config

danlo:
In regards to #468.  I missed that resque-scheduler was *completely* stand-alone and wouldn't catch on I use a different redis database for resque jobs.

I've added this warning in the hopes that other people who also use a different redis database like will avoid this problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,17 @@ default `'text'`)
 * `VERBOSE` - Maximize log verbosity if non-empty (equivalent to a level
 of `MonoLogger::DEBUG`, default `false`)
 
+*Warning: If you use a different redis database, other than the default database (0), to store your redis jobs, you will have store the Resque.redis into the initializer file*
+
+```
+resque-scheduler [other options] --initializer-path script/resque-scheduler.init.rb
+```
+
+```ruby
+# script/resque-scheduler.init.rb
+require 'resque'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -147,6 +147,17 @@ default `'text'`)
 * `VERBOSE` - Maximize log verbosity if non-empty (equivalent to a level
 of `MonoLogger::DEBUG`, default `false`)
 
+*Warning: If you use a different redis database, other than the default database (0), to store your redis jobs, you will have store the Resque.redis into the initializer file*
```
meatballhat: This isn't necessarily true, though.  I think it would be clearer to say that specifying the full Redis URL within the schedule process is important and perhaps give some examples of how to do this, just like in [the resque web example](examples/config/initializers/resque-web.rb).  I would also support adding the Redis URL as an env/cli option so that initializers are not necessary.

juanibiapina:
Does it run standalone even if a dependency to "environment" is specified? In that case it should pick up Redis config from an initializer, right?

The weirdest thing for me is that schedules actually work, but delayed jobs don't.

meatballhat:
@juanibiapina Sorry for delay!  I don't think your question can be answered without knowing the specifics of how your code is loaded.  The flexibility of this part of the configuration is a constant source of question bugs, unfortunately 🙍

juanibiapina:
My problem was another issue. Scheduler wasn't able to elect a master node properly. I cleaned up redis and it worked.

PR #474, opened 2015-04-18
Fix scheduler settings

dtaniwaki:
When I start resque-scheduler for dynamic scheduling, I found the rake task overrides the dynamic setting and disabled it.

I set up resque-scheduler in accordance with the guidance.
https://github.com/resque/resque-scheduler/tree/f0bb556167bd297ef3464e80d6b5b9bde0ab3fdd#rake-integration

Here's the cli setup called after the manual setup.
https://github.com/resque/resque-scheduler/blob/f0bb556167bd297ef3464e80d6b5b9bde0ab3fdd/lib/resque/scheduler/tasks.rb#L17

The cli setup makes hash of all the keys with nil if I don't set it by environment variables.
https://github.com/resque/resque-scheduler/blob/f0bb556167bd297ef3464e80d6b5b9bde0ab3fdd/lib/resque/scheduler/cli.rb#L139

I think we should skip overriding if the value is nil.

[Review on Reviewable](https://reviewable.io/reviews/resque/resque-scheduler/474)

### DIFF
--- a/lib/resque/scheduler/env.rb
+++ b/lib/resque/scheduler/env.rb
@@ -56,24 +56,14 @@ module Resque
 
       def setup_scheduler_configuration
         Resque::Scheduler.configure do |c|
-          if (app_name = options[:app_name]) && !app_name.nil?
-            c.app_name = app_name
+          [:app_name, :env, :logfile, :logformat].each do |sym|
+            if (v = options[sym]) && !v.nil?
+              c.public_send "#{sym}=", v
+            end
           end
 
           if (dynamic = options[:dynamic]) && !dynamic.nil?
             c.dynamic = !!dynamic
-          end
-
-          if (env = options[:env]) && !env.nil?
-            c.env = env
-          end
-
-          if (logfile = options[:logfile]) && !logfile.nil?
-            c.logfile = logfile
-          end
-
-          if (logformat = options[:logformat]) && !logformat.nil?
-            c.logformat = logformat
           end
 
           if (psleep = options[:poll_sleep_amount]) && !psleep.nil?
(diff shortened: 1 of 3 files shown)


lib/resque/scheduler/env.rb
```diff
@@ -56,32 +56,22 @@ def setup_pid_file
 
       def setup_scheduler_configuration
         Resque::Scheduler.configure do |c|
-          if options.key?(:app_name)
```
yaauie: here we're explicitly using a config key IFF it is present (via `Hash#key?`), which is intentional because it allows us to explicitly use nil to unset a configuration variable; your changes make it so we cannot unset these config variables.

dtaniwaki: I understand, but this code set nil for all the keys.
https://github.com/resque/resque-scheduler/blob/f0bb556167bd297ef3464e80d6b5b9bde0ab3fdd/lib/resque/scheduler/cli.rb#L139

yaauie: Then it's probably better to fix it there :smile:

yaauie: Maybe something like this?

~~~ diff
       def options
         @options ||= {}.tap do |o|
-           CLI_OPTIONS_ENV_MAPPING.map { |key, envvar| o[key] = env[envvar] }
+           CLI_OPTIONS_ENV_MAPPING.map { |key, envvar| o[key] = env[envvar] if env.include?(envvar) }
         end
       end
~~~

dtaniwaki: I made another PR
https://github.com/resque/resque-scheduler/pull/475

Thanks. :)

PR #475, opened 2015-04-18
Avoid making keys for nil environment vairables

dtaniwaki:
Moved from https://github.com/resque/resque-scheduler/pull/474

The cli's setup env options has keys for nil values which unset all the configs even after initializer or rake tasks's setup set them.

Could you consider to fix this issue?

[Review on Reviewable](https://reviewable.io/reviews/resque/resque-scheduler/475)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,3 +14,5 @@ Style/DoubleNegation:
   Enabled: false
 Metrics/PerceivedComplexity:
   Enabled: false
+Metrics/ClassLength:
+  Max: 110
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -20,6 +20,7 @@ Resque Scheduler authors
 - David Doan
 - David Yang
 - Denis Yagofarov
+- dtaniwaki
 - Evan Tahler
 - Eugene Batogov
 - Giovanni Cappellotto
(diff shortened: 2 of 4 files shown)


yaauie:
Yeah, I think this is a lot lower surface area & better represents the real issue :smile:

dtaniwaki:
Can I add my name in `AUTHORS` list?
And the lint is a little bit too strict. I can't put any line end in that class nor leave the line that long to pass the tests.

yaauie:
@dtaniwaki every project is different, and while this project tends to be pretty liberal in adding authors to the list (74 of the 95 github users who have committed code are listed), I think I'd leave answering that up to @meatballhat, or one of the more active maintainers.

Feel free to relax the rubocop control for ClassLength a bit; it's supposed to be guidance, but we don't want it getting in the way of a fix.

~~~ diff
diff --git a/.rubocop.yml b/.rubocop.yml
index 197797b..e286351 100644
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,3 +14,5 @@ Style/DoubleNegation:
   Enabled: false
 Metrics/PerceivedComplexity:
   Enabled: false
+Metrics/ClassLength:
+  Max: 110 # default is 100
~~~

In #474 you added a spec to verify the behaviour, but I don't see that here; would you mind adding it (or something similar) to this PR?

meatballhat:
:+1: thank you, @dtaniwaki !  I'm in favor of adding you to `AUTHORS`.  I agree fully with @yaauie's feedback, too :metal: :smiley_cat:

meatballhat:
@dtaniwaki Thank you!

Issue #477, opened 2015-04-30
Blows up when you only use part of Rails


moubry:
When starting `resque-scheduler` via the Rake task, I get this error:

    NoMethodError: undefined method `env' for Rails:Module

https://github.com/resque/resque-scheduler/blob/f0bb556167bd297ef3464e80d6b5b9bde0ab3fdd/lib/resque/scheduler/configuration.rb#L16

I have a pretty barebones Sinatra app—that uses ActionMailer to send emails. I guess ActionMailer is initializing the `Rails` module without `env`.

```irbrc
$ irb
irb(main):001:0> require 'action_mailer'
=> true
irb(main):002:0> Rails
NameError: uninitialized constant Rails
irb(main):003:0> ActionMailer::Base
=> ActionMailer::Base
irb(main):004:0> Rails
=> Rails
irb(main):005:0> Rails.env
NoMethodError: undefined method `env` for Rails:Module
```

yaauie:
@moubry: would you mind submitting a pull-request to fix it? Off the top of my head, something like this could work:

~~~ diff
-        @env ||= Rails.env if defined?(Rails)
+        @env ||= Rails.env if defined?(Rails) && Rails.respond_to?(:env)
~~~

PR #478, opened 2015-04-30
Make sure Rails module responds to :env method

moubry:
Some parts of Rails—such as ActionMailer—will initialize the Rails module without `env` and this line will blow up.

Fixes issue #477.


meatballhat:
Thank you! :heart_eyes_cat:

PR #481, opened 2015-06-15
ActiveJob compatibility

mfo:
Hello ;
I'd like to collect some feedback about this work in progress that makes the scheduler compatible with active job (rails >4.2). 

Actually I'm testing it in production
Actually I'm testing it in production

FIY: 
- I bump the version to scheduler 5
- the readme is not yet up to date
- remove_delayed does not works (and active job used this gem as a queue_adapter for Job.set(wait: tada).


yaauie:
The first issue I see is that this not only makes resque-scheduler *compatible* with ActiveSupport, it makes it *dependent* on ActiveSupport (e.g., requiring the job class to respond to `queue_name`), and I'm pretty sure this is *not* something we would want to take on; this project has many consumers, and some of them (one of my production setups with hundreds of worker processes included) neither have nor want a dependency on ActiveSupport.

mfo:
tx @yaauie I'll check how to avoid that... cheers

meatballhat:
@mfo bump :smiley_cat:

mfo:
@meatballhat bumped ( have not digged how to avoid the dependencies since it's not an issue with my project )

meatballhat:
@mfo OK :+1:

PR #492, opened 2015-09-22
Add "remove_delayed_selection_with_all_job_infos" and "change_delayed_selection_timestamp" methods

guizmaii:
The name of the "remove_delayed_selection_with_all_job_infos" method isn't so great but I can't find a better one for now.

This method is usefull to write more complex matchers. I have this need in one of my project.

The "change_delayed_selection_timestamp" is usefull to change the execution date of a job, matched with a block.


guizmaii:
I need to add tests and documentation ;)

meatballhat:
@guizmaii How about those tests and docs? :smiley_cat:

guizmaii:
Yep, I have to write it but I don't have the time for now. 
I hope to give you this soon.

It's in production at home ;)

meatballhat:
@guizmaii :+1:

codeodor:
Is this basically `remove_delayed_selection` but also providing access to the class name? 

In the readme, I see we can get args ... but can't see how to also do it via class name (unless I have the exact same arguments ... which is not the case for us).

codeodor:
Oh, never mind on my question. I see reading the code for this PR, you can already provide the class as a parameter to that method.

guizmaii:
Hi everyone,

I finally added the tests and the documentation ! 🎉 

Looking forward for your feedback

PR #527, opened 2016-03-29
Race condition in concurrent shutdown/startup (2: lock)

hkdsun:
This is a really scrappy version of the locking method that forces schedulers to wait for others' startup
@fw42


hkdsun:
Oops. Didn't mean to open this PR in this repo. Will submit again later on :)

Issue #528, opened 2016-03-30
Concurrent shutdowns/startups cause corrupted schedules


hkdsun:
We are in the process of integrating resque-scheduler into our product and while monitoring its performance we noticed that some scheduler instances start with corrupted (partial or empty) schedules.

### Problem
The race condition is amplified because in our deploy strategy, there is no guarantee of order regarding when scheduler instances are shutdown/started. After some investigation I have tracked it down to the following sequences:

* Instance 1 receives a TERM signal and starts cleaning the schedule on redis ([using `clean_schedules`](https://github.com/resque/resque-scheduler/blob/master/lib/resque/scheduler.rb#L364))
* Instance 1 is started up again and starts loading the schedules in a transaction [after which it will `reload_schedule!`](https://github.com/resque/resque-scheduler/blob/master/lib/resque/scheduler/scheduling_extensions.rb#L77)
* Instance 2 receives a TERM signal and is blocked by Instance 1 starting up (since startup is done atomically)
* Instance 2 starts cleaning schedules as soon as Instance 1 is done inserting its schedules into redis but before Instance 1 does a `reload_schedule!`, some time [after line 74 but before line 77](https://github.com/resque/resque-scheduler/blob/master/lib/resque/scheduler/scheduling_extensions.rb#L74-L77)
* Instance 1 tries to `reload_schedule!` but loads a corrupt schedule because Instance 2 has removed some of its entries

This causes Instance 1's underlying rufus-scheduler to have a partial/empty schedule for the duration that is up.

I have unit tests that prove this behaviour.

### Solution
We've came up with a good solution that involves removing [a peculiar line](https://github.com/resque/resque-scheduler/blob/master/lib/resque/scheduler.rb#L364) in the scheduler's shutdown hook:
```
      def handle_signals_with_operation
        yield if block_given?
        handle_signals
        false
      rescue Interrupt
        if @shutdown
          Resque.clean_schedules # this can be removed with no obvious impact
          release_master_lock
        end
        true
      end
```

Looking through the git history, this line was added so that redis is cleaned of remainders of resque-scheduler but at the moment [this operation is done during startup, in an atomic fashion](https://github.com/resque/resque-scheduler/blob/master/lib/resque/scheduler/scheduling_extensions.rb#L55-L65), so there is no need to have the schedules cleaned on shut down. Removing this does not break any of the tests.

[...]

PR #531, opened 2016-04-11
Fix race condition in concurrent shutdown/startup

hkdsun:
This PR addresses issue https://github.com/resque/resque-scheduler/issues/528

Digging through the git history it's not clear why we need to `Resque.clean_schedules` on shutdown since it's done when resque-scheduler starts up. Removing it doesn't fail any tests and fixes the race condition described in #528.

This fix has been monitored in Shopify's environment and it has not been cause of any issues

cc @fw42 @sirupsen

### DIFF
--- a/lib/resque/scheduler.rb
+++ b/lib/resque/scheduler.rb
@@ -361,11 +361,12 @@ module Resque
         handle_signals
         false
       rescue Interrupt
-        if @shutdown
-          Resque.clean_schedules
-          release_master_lock
-        end
+        before_shutdown if @shutdown
         true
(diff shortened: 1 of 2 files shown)


hkdsun:
@meatballhat @bugant would be great to get any of your opinions here

fw42:
:+1: Looks good to me, but need to convert to not use keyword arguments (that's why the tests are failing).

Sirupsen:
LGTM

fw42:
maybe @steveklabnik can take a look too?

PR #532, opened 2016-04-12
Fix error on uninitialized poll sleep amount

hkdsun:
**Problem:** latest master cannot run without specifying the `RESQUE_SCHEDULER_INTERVAL` environment variable

**Solution:** we lazily load `@poll_sleep_amount`; so we should not access it directly

Added a unit test that would fail without the fix.

@fw42 @meatballhat

### DIFF
--- a/lib/resque/scheduler.rb
+++ b/lib/resque/scheduler.rb
@@ -336,11 +336,11 @@ module Resque
 
       def poll_sleep_loop
         @sleeping = true
-        if @poll_sleep_amount > 0
+        if poll_sleep_amount > 0
           start = Time.now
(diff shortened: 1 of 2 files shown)


fw42:
@steveklabnik if you have a minute

steveklabnik:
I know next to nothing about how resque-scheduler is structured, but i think this looks okay?

fw42:
@steveklabnik: Any recommendations who else we should ping for review (preferably someone with commit access)?

steveklabnik:
It looks like @meatballhat was the last to merge some PRs, maybe they are around?

PR #534, opened 2016-04-20
Resque::Scheduler.logger should be writable by user

hkdsun:
I imagine it is a very common use-case to swap out loggers between applications. This PR will let the user choose their own logger as long as they adhere to [`Logger`](http://ruby-doc.org/stdlib-2.3.0/libdoc/logger/rdoc/Logger.html). It also changes the corresponding test so that we're not doing private method dances

@fw42 @Sirupsen @meatballhat @bugant @dylanahsmith I'd appreciate some of your 👀  here

### DIFF
--- a/test/scheduler_setup_test.rb
+++ b/test/scheduler_setup_test.rb
@@ -14,8 +14,10 @@ context 'Resque::Scheduler' do
 
   test 'set custom logger' do
     custom_logger = MonoLogger.new('/dev/null')
-    Resque::Scheduler.send(:logger=, custom_logger)
-    assert_equal(custom_logger, Resque::Scheduler.send(:logger))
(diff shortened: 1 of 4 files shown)


Sirupsen:
LGTM

test/scheduler_setup_test.rb
```diff
@@ -14,8 +14,10 @@
 
   test 'set custom logger' do
     custom_logger = MonoLogger.new('/dev/null')
-    Resque::Scheduler.send(:logger=, custom_logger)
```
fw42: 💩

fw42:
LGTM

hkdsun:
Another bonus fix, I just added a fix for the failing CI test

dylanahsmith:
:+1:

PR #535, opened 2016-04-20
Refactor schedule storage for scheduling extension

hkdsun:
With 50+ instances of resque-scheduler all using the same redis we were constantly experiencing the race condition mentioned in this comment: 
https://github.com/resque/resque-scheduler/blob/master/lib/resque/scheduler/scheduling_extensions.rb#L47-L52

The sequence of events that reproduces the race condition is as follows, assuming Redis currently stores S1 (Schedule #1):
* RS_1 gets the list of non_persistent_schedules and gets S1
* RS_2 gets the list of non_persistent_schedules and gets S1
* RS_1 cleans S1 and loads S2
* RS_1 uses S2 as its schedule
* RS_2 cleans S1 (but it's not there) and loads S3
* RS_2 uses S2+S3 as its schedule => :skull:

After a lot of discussion in https://github.com/Shopify/resque-scheduler/pull/7 I refactored how the schedules are stored. The gist of the discussion is that the way we stored schedules in redis was flawed:
* There is no reason to use Redis for non-persistent schedules if we `clean_schedules(non_persistent_schedules)` from redis on every startup anyway
* 100+ instances trying to clean and reload schedules on Redis at the same time every time the scheduler cluster is reloaded creates a lot of contention and leads to unnecessary race conditions that are hard to manage and require hacky solutions
* Redis should be used as the single source of truth only for dynamic+persistent schedules (imo these two should be same thing)
* non-persistent (dynamic or not) schedules should be local to a single instance, if persistence across cluster is desired, then dynamic+persistent schedules should be used (this is why we need to marry the dynamic and persistent options)

This PR essentially separates the storage of persistent schedules and non-persistent schedules. Persistent schedules are now stored in Redis under the `persistent_schedules` key, non-persistent schedules are stored in instance's memory under `@non_persistent_schedules`, and schedule updates are communicated through Redis (`schedules_changed`) as per usual. The public interface however has not been changed. 

This problem has gone through many reviewing iterations in the following PRs before being put into Shopify's production environment; for the curious: https://github.com/Shopify/resque-scheduler/pull/8, https://github.com/Shopify/resque-scheduler/pull/7, https://github.com/Shopify/resque-scheduler/pull/4

Just to give you an idea of the improvement, compare resque-scheduler's reaction to our deploys before the fix:
image
and after:
image

@bugant @meatballhat 
\cc @Sirupsen @fw42 @dylanahsmith


Sirupsen:
This has been running well internally. We'll leave this up for a bit, and maybe @meatballhat has context on why the non-persistent schedules were persisted. We'll merge this in a couple of days otherwise, as this is backwards compatible.

fw42:
I don't like the `SchedulingExtensions` module.. There is too many things happening there. Would it make sense to move a lot of the logic into separate helper classes? Like `PersistentSchedule` and `NonPersistentSchedule` (possibly both subclassing from a `Schedule` class if that makes sense). Both classes could share the same interface. I think that would clean up the module's code considerably.

test/multi_process_test.rb
```diff
@@ -7,36 +7,34 @@
...
-    # doing this outside the threads increases the odds of failure
-    Resque.schedule = schedules
-    counts << Resque.schedule.size
-
-    counts += get_results_from_children(pids)
+    schedules += get_results_from_children(pids)
 
-    counts.each_with_index do |c, i|
-      assert_equal schedule_count, c, "schedule count is incorrect (c: #{i})"
+    schedules.each_with_index do |schedule, i|
```
fw42: Not sure if this makes sense, but what jumps out to me here is that the assertion on the line below might pass (or rather: might never run) if `schedules` is empty. Maybe add a `assert_equal some_number, schedules.size`?

test/scheduler_test.rb
```diff
@@ -313,7 +311,7 @@
     }
     assert_equal(
       { 'cron' => '* * * * *', 'class' => 'SomeIvarJob', 'args' => '/tmp/75' },
-      Resque.decode(Resque.redis.hget(:schedules, 'my_ivar_job'))
+      Resque.schedule['my_ivar_job']
```
fw42: this change is because they are not being persisted anymore?

fw42:
I think this PR looks ok. A bit messy. The code could use some cleanup (the existing code, not the code you are adding). But I think this is a step into the right direction. Feel free to merge, imho.

👍

test/scheduler_test.rb
hkdsun: Yes the previous semantics assumed that redis is the single source of truth thus it looked in redis for schedules

test/multi_process_test.rb
hkdsun: good point :)

meatballhat:
@Drakula2k I completely missed the key rename.  That being said, I don't typically think of the storage scheme as being part of the resque-scheduler API.  How are you depending on the redis key names?

hkdsun:
@Drakula2k In my opinion, access to the schedule has always been done through the [`Resque.schedule` method](https://github.com/resque/resque-scheduler/blob/master/lib/resque/scheduler/scheduling_extensions.rb#L59). This PR only changes the underlying storage for that method which is why the change wasn't mentioned in the version bump

Drakula2k:
Well, I just updated the gem and all my schedules are gone. Pretty unexpected. I had to manually rename the key.

meatballhat:
@Drakula2k @hkdsun more discussion over here, please 😸 : https://github.com/resque/resque-scheduler/issues/563

PR #536, opened 2016-04-21
Add on_failure hooks

hkdsun:
This PR adds an `on_failure` hook so that the user can handle failures themselves. This is very useful especially when all exceptions need to be parsed by the user's application for logging purposes.

@fw42 @dylanahsmith @bugant @Sirupsen @meatballhat
would appreciate any feedback


dylanahsmith:
Doesn't resque already have failure hooks?

fw42:
resque does. This is resque-scheduler. To rescue exceptions that happen when scheduling jobs (not exceptions that are being raised during job execution).

lib/resque/scheduler.rb
```diff
@@ -208,16 +210,24 @@ def enqueue_delayed_items_for_timestamp(timestamp)
...
+      end
+
       def handle_shutdown
         exit if @shutdown
         yield
         exit if @shutdown
       end
 
-      def handle_errors
+      def handle_errors(fallback)
```
fw42: do we really need this method? seems unnecessary now. I would just inline this code in the `enqueue` method.

lib/resque/scheduler.rb
```diff
@@ -208,16 +210,24 @@ def enqueue_delayed_items_for_timestamp(timestamp)
         end
       end
 
+      def enqueue(config)
```
fw42: I would change this method to
```ruby
def enqueue(config)
  enqueue_from_config(config)
rescue => e
  Resque::Scheduler::Plugin.run_failure_hooks(failure_handler, config, e)
end
```
and avoid all the lambda stuff.

lib/resque/scheduler.rb
hkdsun: agreed

lib/resque/scheduler.rb
hkdsun: done

test/scheduler_hooks_test.rb
```diff
@@ -21,4 +21,36 @@
...
+
+  test 'default failure hooks are called when enqueueing a job fails' do
+    config = {
+      'cron' => '* * * * *',
+      'class' => 'SomeRealClass',
+      'args' => '/tmp'
+    }
+
+    e = RuntimeError.new('custom error')
+    Resque::Scheduler.stubs(:enqueue_from_config).raises(e)
```
fw42: `s/stubs/expects/`?

test/scheduler_hooks_test.rb
```diff
@@ -21,4 +21,36 @@
...
+    Resque::Scheduler.failure_handler = ExceptionHandlerClass
+
+    config = {
+      'cron' => '* * * * *',
+      'class' => 'SomeRealClass',
+      'args' => '/tmp'
+    }
+
+    e = RuntimeError.new('custom error')
+    Resque::Scheduler.stubs(:enqueue_from_config).raises(e)
```
fw42: same

test/scheduler_hooks_test.rb
```diff
@@ -21,4 +21,36 @@
...
+
+    Resque::Scheduler.expects(:on_failure_default).with(config, e)
+    Resque::Scheduler.enqueue(config)
+  end
+
+  test 'failure hooks are called when enqueueing a job fails' do
+    Resque::Scheduler.failure_handler = ExceptionHandlerClass
```
fw42: this is going to leak into other tests... should unset it in an `ensure` block to the value it had before.. maybe make a `with_failure_handler` test helper for that?

fw42:
Looks mostly good to me 👍

dylanahsmith:
I don't see why we need another failure hook though.  We don't have a separate failure hook for failures in before_perform, so why should we have one for enqueue hooks when they are enqueued through resque scheduler?

What would you do differently on these hooks rather than the resque hooks?

fw42:
@dylanahsmith: Are you saying that the exceptions that this PR here would catch are already being caught by the Resque failure hooks?

hkdsun:
I doubt that's true but I could be wrong

dylanahsmith:
I'm not saying they would be caught by the resque failure hooks.  But I'm wondering why we need another failure handler (`Resque::Scheduler.failure_handler`).

fw42:
oh, I see, yeah, I'm ok with just plugging in the one from resque

lib/resque/scheduler.rb
```diff
@@ -208,18 +210,18 @@ def enqueue_delayed_items_for_timestamp(timestamp)
...
+        enqueue_from_config(config)
+      rescue => e
+        Resque::Scheduler::Plugin.run_failure_hooks(failure_handler, config, e)
```
dylanahsmith: Why are there plugins on the failure handler?  Can't we just call a method on it?

hkdsun: No strong opinions from me. 

Thoughts on saving the default behaviour (logging to `logger`) and in addition to that having a backend? Is that what resque/resque does? e.g.
```ruby
Resque::Scheduler.failure_handlers.each do |handler|
  handler.on_failure(config, e)
end
```

dylanahsmith: We should keep the default logging behaviour if the backend isn't specified.  I'm not sure if that should be inside the failure handler or not.

dylanahsmith:
By the way, I'm sold on needing the separate failure handling for resque scheduler, since the resque one doesn't seem suitable for re-use here (e.g. it expects a Resque::Worker instance) and the previous behaviour of just logging the error isn't sufficient for notifying developers of the error.

lib/resque/scheduler/failure.rb
```diff
@@ -0,0 +1,16 @@
+module Resque
+  module Scheduler
+    module Failure
+      # allow user to set an additional failure handler
+      attr_writer :failure_handler
+
+      def failure_handler
+        @failure_handler ||= self
```
dylanahsmith: I think you should use a small default FailureHandler class to do this, e.g.

```ruby
class Resque::Schduler::Handler
  def self.on_enqueue_failure(_, e)
    Resque::Schduler.log_error ...
  end
end
```

lib/resque/scheduler/failure.rb
```diff
@@ -0,0 +1,16 @@
+module Resque
+  module Scheduler
+    module Failure
+      # allow user to set an additional failure handler
+      attr_writer :failure_handler
```
dylanahsmith: I think this would be more discoverable if it just lived in Resque::Scheduler, which is the only class that includes this module.  Moving code into a module when it is only included into a single class is an anti-pattern.

dylanahsmith:
Other than a couple of nitpicks about organizing the code for the failure handler, LGTM

hkdsun:
cool i agree with and took care of your comments. thanks for reviewing

dylanahsmith:
:+1:

PR #537, opened 2016-04-25
Make logger accessible

hkdsun:
Now that we have a failure handler/hook, we should allow the user to freely choose how they make a call to the logger. 

Additionally, if the user uses a custom logger by doing the following on initialization
```ruby
Resque::Scheduler.logger = CustomLogging.logger
```
Then they should be able to do the following in their failure hook.
```ruby
Resque::Scheduler.logger.info("msg", custom_param: "test")
```
This PR makes the logger attribute accessible to the user.

@fw42 @Sirupsen @dylanahsmith

### DIFF
--- a/lib/resque/scheduler.rb
+++ b/lib/resque/scheduler.rb
@@ -23,10 +23,10 @@ module Resque
     public
 
     class << self
+      attr_writer :logger
+
       # the Rufus::Scheduler jobs that are scheduled
       attr_reader :scheduled_jobs
-
-      attr_writer :logger
 
       # allow user to set an additional failure handler
       attr_writer :failure_handler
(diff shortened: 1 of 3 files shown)


lib/resque/scheduler.rb
```diff
@@ -23,11 +23,11 @@ module Scheduler
     public
 
     class << self
+      attr_accessor :logger
```
dylanahsmith: Continue using `attr_writer :logger` since you are already defining the method for the reader below

hkdsun: done.

dylanahsmith:
:+1:

fw42:
LGTM. Would maybe add a test. But don't feel strongly. 👍

hkdsun:
Added a test. It should make the intention clear for a future maintainer

fw42:
Cool 👍

PR #539, opened 2016-04-29
version: bump to 4.2.0

Sirupsen:
FYI @meatballhat


Sirupsen:
FYI history.md was updated too https://github.com/resque/resque-scheduler/blob/master/HISTORY.md#420-2016-04-29

PR #542, opened 2016-05-11
Fix registering signal keys for Windows.

irasally:
Some singals(USR,1 USR2, QUIT) are not exist in Windows.
So when registering signal keys, we need to check system signal list.


meatballhat:
@irasally Thank you!

PR #545, opened 2016-05-19
improve find_delayed_selection performance

crazyshot:
instead of pulling job-by-job with separate requests from redis,
pull everything in single range and work in memory only later

for 1000 queued jobs, improves about 20 times

### DIFF
--- a/lib/resque/scheduler/delaying_extensions.rb
+++ b/lib/resque/scheduler/delaying_extensions.rb
@@ -188,15 +188,20 @@ module Resque
 
         timestamps = redis.zrange(:delayed_queue_schedule, 0, -1)
 
-        jobs_per_ts = redis.pipelined do |r|
-          timestamps.each do |ts|
-            r.lrange("delayed:#{ts}", 0, -1)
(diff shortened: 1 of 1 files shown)


carsonreinke:
This is great, but shouldn't the list of timestamps be batched instead because memory consumption could be quite high.

crazyshot:
For amount that would fill your memory, previous version of code would work forever :)
Theoretically you're right, but only in case where there are a lot of jobs with HUGE payload, for normal use, even 100k jobs will take few MB, since it's just strings and parsed payload is discarded after each job test.

I can refactor it to use each_slice(N) on timestamps with another flatten in the end, but not sure what N should be, there's no way to know what size of payloads might be.

crazyshot:
I also thought about adding "uniq" there before "select", will improve in case of many duplicate jobs, but then people might expect to get each of duplicate jobs in the filter block.

crazyshot:
rebased on new master

meatballhat:
Lovely! 😻

PR #564, opened 2016-07-05
Easier to run tests by using Docker

carsonreinke:
Pretty easy to run some tests locally and having Redis running for you.

```
$> docker build -t resque-scheduler .
$> docker run resque-scheduler
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+#docker build -t resque-scheduler .
+#docker run resque-scheduler
+
+FROM ubuntu:trusty
+
+RUN apt-get -y update
+RUN apt-get -y install redis-server
+RUN apt-get -y install curl
+RUN apt-get -y install git patch  gawk g++ gcc make libc6-dev patch libreadline6-dev zlib1g-dev libssl-dev libyaml-dev libsqlite3-dev sqlite3 autoconf libgmp-dev libgdbm-dev libncurses5-dev automake libtool bison pkg-config libffi-dev
+
+#Ruby
+RUN gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+#Project files
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+COPY . /usr/src/app
+
+#Install gems
+RUN ["/bin/bash", "-l", "-c", "cd /usr/src/app; bundle install"]
+
+#Run Redis and tests
+CMD ["/bin/bash", "-l", "-c", "redis-server > /dev/null & cd /usr/src/app && bundle exec rake test"]
```
meatballhat: I would prefer to either make this Dockerfile more friendly for general use, or move it inside `./test/` with some instructions in the `README` on how it may be used.  If we're going to have *both*, then I think a general use top-level Dockerfile should:

- use `FROM ruby:alpine`
- assume a redis configuration is going to be injected
- (trivial) work/install/run in `/app` instead of `/usr/src/app`
- use `ENTRYPOINT ["bundle", "exec", "resque-scheduler"]` and `CMD ["--"]` (no-op)

meatballhat:
In addition to the file comments, I should also say that I think shipping an "official" docker container to hub.docker.com would be a nice addition 😸

Dockerfile
carsonreinke: Moving to `./test/` is not possible with Docker, since the `COPY` is not allow to go above the current directory.

Regarding the `ENTRYPOINT`, `CMD` is required to start the local Redis server, unless there is something I am missing.

carsonreinke:
Yeah, that would be nice to have.  I would assume ENV would be the way to communicate most of the settings (Redis, YAML file, etc).

I will make these corrections.  Just toying with this idea, so don't feel obligated to merge if this will muddle with useless files.

PR #568, opened 2016-07-17
Default quiet=false when background with logfile

### LINKED ISSUE #565: Logging quiet when backgrounded even when logfile is set
Resque log is always empty after some recent updates. May be related to https://github.com/resque/resque-scheduler/pull/559

I use `resque-scheduler` and `capistrano-resque`. Parts of my configuration:

_resque.rake_:

``` ruby
  task :setup_schedule => :setup do
    ENV['VERBOSE'] = '1'
    ENV['QUIET'] = nil
    ENV['LOGFILE'] = Rails.root.join("log", "resque-scheduler.log").to_s
```

_deploy.rb_

``` ruby
set :resque_verbose, true
set :resque_extra_env, "VERBOSE=1"
```

So I tried to enable `VERBOSE` even in different places and it still doesn't work.

This is how capistrano starts scheduler:

```
$ cap production resque:scheduler:start
00:00 resque:scheduler:start
      01 nohup bundle exec rake RAILS_ENV=production PIDFILE=/…/shared/tmp/pids/scheduler.pid BACKGROUND=yes VERBOSE=1 MUTE=1 resque:scheduler >> /dev/null 2>> /dev/null
```

meatballhat:
Closes #565

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Fixed
 - Reporting version via `resque-scheduler --version`
+- Verbosity when run in background with logfile present
 
 ## [4.3.0] - 2016-06-26
 ### Added
--- a/lib/resque/scheduler/env.rb
+++ b/lib/resque/scheduler/env.rb
@@ -30,13 +30,7 @@ module Resque
       def setup_backgrounding
         return unless options[:background]
 
-        # Need to set this here for conditional Process.daemon redirect of
-        # stderr/stdout to /dev/null
-        Resque::Scheduler.quiet = if options.key?(:quiet)
-                                    !!options[:quiet]
-                                  else
-                                    true
-                                  end
+        Resque::Scheduler.quiet = background_quiet?
 
         unless Process.respond_to?('daemon')
           abort 'background option is set, which requires ruby >= 1.9'
@@ -44,6 +38,12 @@ module Resque
 
         Process.daemon(true, !Resque::Scheduler.quiet)
         Resque.redis.client.reconnect
+      end
+
+      def background_quiet?
+        return !!options[:quiet] if options.key?(:quiet)
+        return false if options.key?(:logfile)
+        !!options[:background]
       end
 
       def setup_pid_file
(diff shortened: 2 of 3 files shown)


meatballhat:
@xeron How's this? 😸

xeron:
Well, it may help, but the whole situation raises some questions.

* Why do you have different behavior in background/foreground mode?
* Why do you have 2 variables controlling logging – `VERBOSE` and `QUIET`?
  * It's confusing
  * Seems like you don't respect `VERBOSE` in background mode? Or in both bg/fg modes?

I see this code in logger builder:
```ruby
      def level
        if @verbose && !@quiet
          MonoLogger::DEBUG
        elsif !@quiet
          MonoLogger::INFO
        else
          MonoLogger::FATAL
        end
      end
```

but looks like in background mode you're using some different logic?

meatballhat:
@xeron I agree that the current situation with `verbose` and `quiet` is confusing.  Legacy code!  (and some recent changes)

My understanding of the current intent, is: 
``` ruby
default_quiet = false

if options[:background]
  default_quiet = true
end

if options[:logfile]
  default_quiet = false
end

log_level = if options.fetch(:quiet, default_quiet)
              FATAL
            elsif options.fetch(:verbose, false)
              DEBUG
            else
              INFO
            end
```

I'm very much ready to believe that the current implementation does not match this intent.  Supporting explicit log levels instead of the `verbose` / `quiet` thing sounds great to me, assuming that's what you're suggesting 😸

xeron:
Yeah, I think keeping only `logfile` and `loglevel` is better. But this is up to you. With explanation of what `QUIET` / `VERBOSE` actually do it's a bit cleaner for me now (if it works as you described).

xeron:
Soo, what happened to this PR and issue I reported? This is still affect me :(

bolsen:
I started using resque-scheduler and am noticing the same issue that @xeron is reporting here. Is there a workaround at least so we can get logs?

PatrickTulskie:
@meatballhat hey just a heads up, I'm making moves to get through a bunch of the PR backlog for resque scheduler. This looks like a good candidate for 5.0.1. If you have a chance, can you rebase and push this up again and I'll get it live? If you're not or I don't hear back from you in the next week or so, I'll have my agent rebase and open a fresh PR with your change so you still get credit for the contribution.

PR #574, opened 2016-08-05
factor out recurring enqueue block

jpittis:
## What does this change?
It factors out the block `resque-scheduler` passes to `rufus-scheduler` into a separate method. This PR makes no logic changes.

## Why make the change?
We use thread local state to determine which redis instance to enqueue a job to. `rufus-scheduler` calls this block from within a new thread. This refactor provides a method we can easily monkey patch so that we can set appropriate thread local state after the new thread has been created.

@fw42 @dylanahsmith 
cc: @Sirupsen

### DIFF
--- a/lib/resque/scheduler.rb
+++ b/lib/resque/scheduler.rb
@@ -141,11 +141,7 @@ module Resque
             args = [args, nil, job: true] if args.is_a?(::String)
 
             job = rufus_scheduler.send(interval_type, *args) do
-              if master?
-                log! "queueing #{config['class']} (#{name})"
-                Resque.last_enqueued_at(name, Time.now.to_s)
-                enqueue(config)
-              end
+              enqueue_recurring(name, config)
             end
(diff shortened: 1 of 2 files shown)


lib/resque/scheduler.rb
```diff
@@ -161,6 +157,14 @@ def load_schedule_job(name, config)
         end
       end
 
+      def enqueue_recurring(name, config)
```
dylanahsmith: This method should be private

dylanahsmith:
```
Offenses:
lib/resque/scheduler.rb:11:3: C: Module has too many lines. [317/314]
  module Scheduler ...
  ^^^^^^^^^^^^^^^^
```
Looks like you need to modify Metrics/ModuleLength in .rubocop_todo.yml

It also looks like a rebase is needed

fw42:
👍

dylanahsmith:
:+1:

PR #578, opened 2016-09-02
Migration task

joshuacronemeyer:
Hi there! Thanks for resque-scheduler it's a really useful library! 

I recently migrated resque into a standalone redis DB and I used this rake task to migrate the resque-scheduler jobs. I thought it might be useful to others so I'm making this pull request for your consideration.  Let me know if you think this utility belongs in the codebase.  

If you think it does belong then I'm happy to make changes and improvements. Here are some things I thought we might want to discuss:
- Speed: I chose to use the scheduler API rather than manipulating the underlying redis data to keep the code cleaner and easier to maintain. It migrated 300 jobs on my staging server in about 90 seconds, which is slower than I liked.
- Safety: Current implementation does nothing to protect you if something goes wrong during processing. If the command were to fail in the middle of a migration you'd have to clean up the mess manually (although it leaves the old data intact so you can try again).

Anyway, let me know if you think this utility is something you'd like in your codebase. If so we can discuss the issues above and any others you'd care to consider. I notice that redis has a MIGRATE command http://redis.io/commands/MIGRATE that supports bulk operations and does pipelining and makes some nice guarantees. Probably is the way to go if we care to fix the issues i raised above.

Thanks again!

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -35,6 +35,7 @@ Resque Scheduler authors
 - Jonathan Conley
 - Jonathan Hyman
 - Jonathan Owens
+- Josh Cronemeyer
 - Joshua Szmajda
 - Justin Weiss
 - Les Hill
--- a/README.md
+++ b/README.md
@@ -640,7 +640,9 @@ on a dedicated Redis database.  While making and managing scheduled tasks,
 `resque-scheduler` currently scans the entire Redis keyspace, which may cause
 latency and stability issues if `resque-scheduler` is hosted on a Redis instance
 storing a large number of keys (such as those written by a different system
-hosted on the same Redis instance).
+hosted on the same Redis instance). The resque:migrate_scheduler rake task is
+provided to move scheduled jobs to a new redis instance when changing your
+application's redis configuration.
 
 #### Compatibility Notes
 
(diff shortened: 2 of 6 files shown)


joshuacronemeyer:
Update: I have researched further using the migrate command to make this faster and find major roadblocks. First: the redis MIGRATE command doesn't support password protected servers!!! (i know, right?) Second: the MIGRATE command doesn't work with different DB's on the same instance so we'd need to run 2 redis instances in order to run the tests for it. Third: the redis namespace gem doesn't support the MIGRATE command.

For now in order to speed things up i'll look into just using regular redis commands and connecting to the other instance the same way i already do. Benefit is that we'll get a speedup. Drawback but won't get some of the safety. But it should work!

joshuacronemeyer:
Ok, last update here unless I get some feedback from you guys. I went ahead and addressed the speed issues as best as i could.

I Used this tool on our production servers tonight and it worked just fine. Migration time for 1300 scheduled tasks was about 12 seconds.

iloveitaly:
@joshuacronemeyer I don't think this is the sort of thing that belongs in the core codebase. Maybe you could post this as a gist and submit a PR to link the gist to the readme here?

PR #580, opened 2016-09-06
Remove misleading text in README

stevenjackson:
As of 4.2 the store for non-persisted schedules was recently changed from redis to in-memory. Resque web has no way to access the actual schedules unless you load them separately in each process.  So users need to know they must load them for resque-web.


carsonreinke:
@meatballhat can we merge this?

mcjoejoe:
Please accept this and merge it.  I spent much too long the other day getting introduced to resque-scheduler trying to figure out why the schedule wasn't showing on the web panel...

PR #588, opened 2016-11-17
Refresh script sha if it doesnt exist on Redis server

hkdsun:
For the `Resilient` lock, we currently raise `Redis::CommandError` when a lua script disappears in the Redis server for some reason. This can happen when a DBA needs to run [`SCRIPT FLUSH`](http://redis.io/commands/script-flush) or when Redis is wiped before each unit test in a test suite.  

This PR lets resque-scheduler try to re-acquire the sha that was lost.

@Sirupsen @fw42 
cc @wvanbergen @kirs

### DIFF
--- a/lib/resque/scheduler/lock/resilient.rb
+++ b/lib/resque/scheduler/lock/resilient.rb
@@ -32,7 +32,7 @@ module Resque
 
         private
 
-        def evalsha(script, keys:, argv:, refresh: false, final_try: false)
+        def evalsha(script, keys:, argv:, refresh: false)
           sha_method_name = "#{script}_sha"
           Resque.redis.evalsha(
             send(sha_method_name, refresh),
@@ -40,8 +40,9 @@ module Resque
             argv: argv
           )
         rescue Redis::CommandError => e
-          if e.message =~ /NOSCRIPT/ && !final_try
-            evalsha(script, keys: keys, argv: argv, refresh: true, final_try: true)
+          if e.message =~ /NOSCRIPT/
+            refresh = true
+            retry
           else
             raise
           end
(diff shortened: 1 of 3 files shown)


lib/resque/scheduler/lock/resilient.rb
```diff
@@ -6,16 +6,17 @@ module Scheduler
...
       class Resilient < Base
         def acquire!
-          Resque.redis.evalsha(
-            acquire_sha,
+          evalsha(
+            :acquire,
             keys: [key],
             argv: [value]
           ).to_i == 1
+        rescue
```
kirs: Are you sure we need this line? Maybe at least log it.

lib/resque/scheduler/lock/resilient.rb
```diff
@@ -32,6 +33,21 @@ def timeout=(seconds)
...
+        def evalsha(script, keys:, argv:, refresh: false, final_try: false)
+          sha_method_name = "#{script}_sha"
+          Resque.redis.evalsha(
+            send(sha_method_name, refresh),
+            keys: keys,
+            argv: argv
+          )
+        rescue Redis::CommandError => e
+          if e.message =~ /NOSCRIPT/ && !final_try
+            evalsha(script, keys: keys, argv: argv, refresh: true, final_try: true)
```
fw42: I think instead of calling the method recursively (and using this `final_try` flag), it might be cleaner to use the `retry` keyword (and set `refresh = true` before calling it).

lib/resque/scheduler/lock/resilient.rb
hkdsun: sorry i dont know how this got here

lib/resque/scheduler/lock/resilient.rb
hkdsun: Good idea. Done

wvanbergen:
:heart::heart::heart::heart::heart::heart:

hkdsun:
We've been running this change on production at Shopify, gone through many deploys, and there's been no issues at all

PR #592, opened 2016-12-15
Gracefully shut down rufus scheduler threads

hkdsun:
## Problem
Rufus scheduler is used as the engine for the scheduling extension of resque-scheduler (recurring jobs). 

When resque scheduler is instructed to shut down (`INT`/`TERM`), it doesn't gracefully shut down its rufus scheduler threads which can lead to surprising results.

For example, if a job has a `before_enqueue` hook that can take a long time to complete (e.g. talks to a slow consumer), then we can have situations where the hook is ran but the job is never enqueued – because the rufus scheduler threads were terminated while still running `before_enqueue` hooks.

## Solution
Before shutting down resque scheduler:

* Call `rufus_scheduler.shutdown(:wait)` to stop further jobs being scheduled by rufus scheduler and join all the rufus scheduler worker threads
* Call `rufus_scheduler.join` to wait for rufus scheduler to exit before shutting down

---

I've included a test reproducing the issue. The test is `sleep`-dependent. I would rewrite the test if anyone has good suggestions.

@fw42 @pushrax @dylanahsmith @sirupsen @meatballhat

### DIFF
--- a/test/scheduler_test.rb
+++ b/test/scheduler_test.rb
@@ -12,6 +12,7 @@ context 'Resque::Scheduler' do
     Resque.redis.flushall
     Resque::Scheduler.clear_schedule!
     Resque::Scheduler.send(:instance_variable_set, :@scheduled_jobs, {})
+    Resque::Scheduler.send(:instance_variable_set, :@shutdown, false)
   end
 
   test 'enqueue constantizes' do
@@ -454,6 +455,63 @@ context 'Resque::Scheduler' do
     Resque::Scheduler.env = nil
     assert Resque::Scheduler.send(:build_procline, 'cage') =~
            /#{Resque::Scheduler.send(:internal_name)}: cage/
+  end
+
+  test 'gracefully shuts down rufus-scheduler threads' do
+    if RUBY_ENGINE == 'jruby' || RUBY_PLATFORM =~ /mingw|windows/i
+      omit("forking is not supported on #{RUBY_ENGINE}/#{RUBY_PLATFORM} but " \
+           'this behaviour is best tested using forks')
+    end
+
+    class BeforeEnqueueJob
+      @queue = :quick
+
+      class << self
+        def before_enqueue_example(*)
+          return false if enqueue_started?
+          enqueue_started!
+
+          sleep 5
+          true
+        end
+
+        def enqueue_started?
+          Resque.redis.get('before_enqueue_job:enqueued') == 'true'
+        end
+
+        def perform(*)
+        end
+
+        private
+
+        def enqueue_started!
+          Resque.redis.set('before_enqueue_job:enqueued', 'true')
+        end
+      end
+    end
+
+    schedule = {
+      'BeforeEnqueueJob' => { cron: '* * * * * *', class: 'BeforeEnqueueJob' }
+    }
+
+    pid = fork do
+      Resque::Scheduler.clear_schedule!
+      Resque.schedule = schedule
+      Resque::Scheduler.run
+    end
+
+    begin
+      30.times do
+        break if BeforeEnqueueJob.enqueue_started?
+        sleep 0.1
+      end
+    ensure
+      Process.kill('TERM', pid)
+      Process.wait(pid)
+    end
+
+    assert BeforeEnqueueJob.enqueue_started?, "Job enqueue didn't start in time"
+    assert_equal 1, Resque.size('quick')
   end
 
   context 'printing schedule' do
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -39,7 +39,7 @@ Metrics/MethodLength:
 # Offense count: 2
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 317
+  Max: 322
 
 # Offense count: 1
 Style/CaseEquality:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,6 +5,7 @@ environment:
   RUBYOPT: '-W0'
   COVERAGE: 1
   JRUBY_OPTS: ''
+  TZ: 'UTC'
 
 install:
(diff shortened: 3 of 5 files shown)


test/scheduler_test.rb
```diff
@@ -456,6 +456,52 @@
            /#{Resque::Scheduler.send(:internal_name)}: cage/
   end
 
+  test 'gracefully shuts down rufus-scheduler threads' do
+    class BeforeEnqueueJob
+      @queue = :quick
+
+      def self.before_enqueue_example(*)
+        if Resque.redis.setnx('before_enqueue_job:enqueued', 'true')
```
fw42: hmm I think the intent here could be a bit more clear to make the test easier to understand. How about something like this:
```ruby
class << self
  def before_enqueue_example(*)
    job_has_already_been_enqueued = set_before_enqueue_flag(true)
    return false if job_has_already_been_enqueued
    sleep 5
    true
  end

  private

  def set_before_enqueue_flag(value)
    Resque.redis.setnx('before_enqueue_job:enqueued', value)
  end
end
```

test/scheduler_test.rb
```diff
@@ -456,6 +456,52 @@
...
+
+    pid = fork do
+      Resque.schedule = schedule
+      Resque::Scheduler.run
+    end
+
+    begin
+      Timeout.timeout(3) do
+        loop do
+          break if Resque.redis.get('before_enqueue_job:enqueued') == 'true'
```
fw42: not a huge fan of how this uses the internals of the job.. maybe there could be a `BeforeEnqueueJob.enqueue_started?` helper method? that would make the test easier maybe

test/scheduler_test.rb
```diff
@@ -456,6 +456,52 @@
            /#{Resque::Scheduler.send(:internal_name)}: cage/
   end
 
+  test 'gracefully shuts down rufus-scheduler threads' do
+    class BeforeEnqueueJob
+      @queue = :quick
+
+      def self.before_enqueue_example(*)
+        if Resque.redis.setnx('before_enqueue_job:enqueued', 'true')
+          sleep 5 # sleep longer than test timeout
```
fw42: Hmm I think I would use a class variable here to know when to continue instead of relying on the timeout. Below, you could do something like:
```ruby
pid = fork do
  trap("TERM") { BeforeEnqueueJob.received_shutdown_signal = true }
  Resque.schedule = schedule
  Resque::Scheduler.run
end
```
and then here do `sleep 0.1 until received_shutdown_signal`.

Not sure if that's much better but it seems less flaky and less time-dependent.

test/scheduler_test.rb
```diff
@@ -456,6 +456,52 @@
...
+
+    Resque::Scheduler.enqueue_from_config(schedule['BeforeEnqueueJob'])
+
+    pid = fork do
+      Resque.schedule = schedule
+      Resque::Scheduler.run
+    end
+
+    begin
+      Timeout.timeout(3) do
```
fw42: Using `timeout` here is unnecessary since nothing is blocking. I would do something like this:
```ruby

(3 * 10).times do
  break if BeforeEnqueueJob.enqueue_started?
  sleep 0.1
end

assert BeforeEnqueueJob.enqueue_started?, "Job enqueue didn't start in time"

Process.kill()
Process.wait()
# ...
```
Would that work?

test/scheduler_test.rb
```diff
@@ -456,6 +456,52 @@
...
+        loop do
+          break if Resque.redis.get('before_enqueue_job:enqueued') == 'true'
+          sleep 0.1
+        end
+      end
+    rescue Timeout::Error
+      flunk 'Child process did not terminate in time'
```
hkdsun: This is bad wording. It should be "Job took to long to be enqueued". Will fix

fw42: This message is confusing, since the "Child process" (identified by `pid`) is still running at this point. This timeout is about the before_enqueue hook not having run yet (even if this timeout doesn't get hit, the child is still running).

test/scheduler_test.rb
fw42: Hmm you might be right. for some reason I thought you could define multiple signal handlers...

fw42: could use a different signal (and send both) but I guess that will become a bit messy

fw42: Hormoz suggested to use a pipe between parent and child and use that for signaling... that might work too.

test/scheduler_test.rb
```diff
@@ -456,6 +456,52 @@
...
+    ensure
+      Process.kill('TERM', pid)
+      Process.wait(pid)
+    end
+
+    assert_equal 'true', Resque.redis.get('before_enqueue_job:enqueued')
+    assert_equal 1, Resque.size('quick')
+  end
```
dylanahsmith: I tried reverting the lib changes and running this test and it passed.  It looks like you aren't properly checking that the job actually gracefully shutdown.  Maybe you need to do something in perform and assert that it was done at the end of the test

test/scheduler_test.rb
```diff
@@ -456,6 +456,59 @@
...
+          Resque.redis.get('before_enqueue_job:enqueued') == "true"
+        end
+
+        def perform(*)
+        end
+
+        private
+
+        def set_before_enqueue_flag(value)
+          Resque.redis.setnx('before_enqueue_job:enqueued', value)
```
fw42: What do you think about just using `set` instead of `setnx` here since that's easier to understand and then above you can clean it up a bit more by doing something like this:
```ruby
return false if enqueue_started?
enqueue_started!
...
```

hkdsun:
I followed instructions in https://github.com/jmettraux/rufus-scheduler#i-get-zotimerb41in-initialize-cannot-determine-timezone-from-nil to correct the errors on the AppVeyor CI.

PR #593, opened 2016-12-16
Add test for ensuring graceful shutdown

hkdsun:
Adds a test for reproducing the symptoms seen https://github.com/resque/resque-scheduler/pull/592 but for the delaying extension (i.e. jobs scheduled in the future / ActiveJob retries).

@meatballhat @fw42 @sirupsen @carsonreinke

### DIFF
--- a/test/scheduler_test.rb
+++ b/test/scheduler_test.rb
@@ -457,41 +457,41 @@ context 'Resque::Scheduler' do
            /#{Resque::Scheduler.send(:internal_name)}: cage/
   end
 
+  test 'finishes enqueueing delayed jobs before shutting down' do
+    if RUBY_ENGINE == 'jruby' || RUBY_PLATFORM =~ /mingw|windows/i
+      omit("forking is not supported on #{RUBY_ENGINE}/#{RUBY_PLATFORM} but " \
+           'this behaviour is best tested using forks')
+    end
+
+    t = Time.now + 1 # in the future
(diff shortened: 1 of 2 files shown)


test/scheduler_test.rb
```diff
@@ -457,41 +457,41 @@
...
-
-      class << self
-        def before_enqueue_example(*)
-          return false if enqueue_started?
-          enqueue_started!
-
-          sleep 5
-          true
-        end
+    t = Time.now + 1 # in the future
```
fw42: this seems kinda racy without `Timecope.freeze`, doesn't it? if it takes more than a second to get to the codepath that looks at this timestamp then this might not be "in the future" anymore. At least use something that's more likely to actually be in the future (like an hour instead of a second)?

hkdsun: I think this is fine. The delaying extension would look at any timestamp in the redis timestamp that's less than `Time.now`. The `+ 1` part is just so that `Resque.enqueue_at` puts the job in the delayed queue rather than the actual job queue.

Increasing this time would increase how long the test takes to run as well.

fw42: > The + 1 part is just so that Resque.enqueue_at puts the job in the delayed queue 

what if by the time `enqueue_at` runs, `t` is not smaller than the (then) `Time.now` anymore? I admit that seems unlikely, but possible.

hkdsun: 🤔  I see what you mean.

PR #596, opened 2017-01-29
Add Resque.remove_delayed_with_queue method

thorn:
Currently one cannot remove a job that was queued with `Resque.enque_at_with_queue` method if queue name does not match class's default queue.

### DIFF
--- a/lib/resque/scheduler/delaying_extensions.rb
+++ b/lib/resque/scheduler/delaying_extensions.rb
@@ -151,6 +151,12 @@ module Resque
         end
       end
 
+      # Given an encoded item, remove it from the queue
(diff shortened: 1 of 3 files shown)


lib/resque/scheduler/delaying_extensions.rb
```diff
@@ -151,6 +151,12 @@ def enqueue_delayed(klass, *args)
         end
       end
 
+      # Given an encoded item, remove it from the queue
```
iloveitaly, line 154: Could you improve this comment to clearly indicate that the magic here is when the default queue doesn't match the queue that the job is in?

thorn:
Closing this as similar method was implemented already

PR #599, opened 2017-02-13
Call encode once to remove the possibility of different encoded job values.

kamaljoshi:
This fixes the non cleanup of Resque scheduled sets where the encoded job in the scheduled list and job set were different. In our setup because of the encoded keys being different, there were orphan sets in Redis for which job had already been processed. It led to Redis memory consumption increasing continuously. 

A simple example to reproduce this would be to pass the foo object(from the class described below) as the argument to resque job. 
```
class Foo
  def as_json _options = {}
    {
       generated_at: Time.now.to_f
    }
  end
end

foo = Foo.new
Resque.encode foo
```

### DIFF
--- a/lib/resque/scheduler/delaying_extensions.rb
+++ b/lib/resque/scheduler/delaying_extensions.rb
@@ -62,11 +62,13 @@ module Resque
       # if O(log(n)).  Returns true if it's the first job to be scheduled at
       # that time, else false
       def delayed_push(timestamp, item)
+        encoded_job = encode(item)
+
         # First add this item to the list for this timestamp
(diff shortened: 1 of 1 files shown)


PatrickTulskie:
@kamaljoshi hey just a heads up, I'm making moves to get through a bunch of the PR backlog for resque scheduler. This looks like a good candidate for 5.0.1. If you have a chance, can you rebase and push this up again and I'll get it live? If you're not or I don't hear back from you in the next week or so, I'll have my agent rebase and open a fresh PR with your change so you still get credit for the contribution.

PR #601, opened 2017-02-27
Bugfix support overlap option

### LINKED ISSUE #504: rufus scheduler "overlap" option not working
I noticed this setting in rufus scheduler.  Is it possible to use with resque scheduler?
https://github.com/jmettraux/rufus-scheduler#overlap--false

I tried this

```
  name = "Blah"
  config = {}
  config[:class] = 'TheJob'
  config[:args] = 'arg'
  config[:every] = ['1m', {first_in: 1.minutes}]
  config[:persist] = true
  config[:queue] = 'queue'
  config[:overlap] = false
  Resque.set_schedule(name, config)
```

But it did't work...

peret:
Fixes #504 for recurring jobs. I'm not sure if the behavior should be the same for delayed jobs. Can be added easily, though.

### DIFF
--- a/lib/resque/scheduler.rb
+++ b/lib/resque/scheduler.rb
@@ -421,9 +421,25 @@ module Resque
 
       def enqueue_recurring(name, config)
         if master?
-          log! "queueing #{config['class']} (#{name})"
-          Resque.last_enqueued_at(name, Time.now.to_s)
-          enqueue(config)
+          allow_overlap = (config[:overlap] != false && config['overlap'] != false)
+          klass_name = config['class'] || config[:class]
+          begin
+            klass = Resque::Scheduler::Util.constantize(klass_name)
+          rescue NameError
+            klass = klass_name
+          end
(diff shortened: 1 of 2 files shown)


lib/resque/scheduler.rb
```diff
@@ -419,11 +410,43 @@ def logger
...
+        [klass, klass_name, queue_name]
+      end
+
+      def in_progress?(queue_name)
+        currently_processing = Resque::Worker.working.map(&:job).any? do |job|
+          job['queue'] == queue_name.to_s
+        end
+        return true if currently_processing
+
+        Resque.peek(queue_name, 0, 5000).any?
```
carsonreinke: Whats this jazz about?

peret: Trying to figure out, if there's either a worker still working on the scheduled job or there's already one in the queue.
Granted, it's awkward. I updated the PR.

Issue #600, opened 2017-02-21
Deprecation warning on startup: "Passing command to redis as is"


jaredbeck:
Deprecation warnings on startup:

```
13:07:00 scheduler.1 | Passing 'info' command to redis as is; administrative commands cannot be effectively namespaced and should be called on the redis connection directly; passthrough has been deprecated and will be removed in redis-namespace 2.0 (at /Users/jared/.rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/gems/resque-1.27.2/lib/resque/data_store.rb:60:in `method_missing')
13:07:00 scheduler.1 | Passing 'script' command to redis as is; administrative commands cannot be effectively namespaced and should be called on the redis connection directly; passthrough has been deprecated and will be removed in redis-namespace 2.0 (at /Users/jared/.rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/gems/resque-1.27.2/lib/resque/data_store.rb:60:in `method_missing')
13:07:00 scheduler.1 | Passing 'script' command to redis as is; administrative commands cannot be effectively namespaced and should be called on the redis connection directly; passthrough has been deprecated and will be removed in redis-namespace 2.0 (at /Users/jared/.rbenv/versions/2.2.5/lib/ruby/gems/2.2.0/gems/resque-1.27.2/lib/resque/data_store.rb:60:in `method_missing')
```

The timestamp and process identifier (`scheduler.1`) come from foreman.

Then, on shutdown,

```
13:09:40 scheduler.1 | resque-scheduler: [INFO] 2017-02-21T13:09:40-05:00: Got INT signal
13:09:40 scheduler.1 | resque-scheduler: [INFO] 2017-02-21T13:09:40-05:00: Shutting down
13:09:40 scheduler.1 | rake aborted!
13:09:40 scheduler.1 | Redis::CommandError: NOSCRIPT No matching script. Please use EVAL.
13:09:40 scheduler.1 | /Users/jared/.rbenv/versions/2.2.5/bin/bundle:23:in `load'
13:09:40 scheduler.1 | /Users/jared/.rbenv/versions/2.2.5/bin/bundle:23:in `<main>'
13:09:40 scheduler.1 | Interrupt: Interrupt
13:09:40 scheduler.1 | /Users/jared/.rbenv/versions/2.2.5/bin/bundle:23:in `load'
[...]

PR #602, opened 2017-03-05
Fix redis-namespace deprecations

carsonreinke:
Fixes part of https://github.com/resque/resque-scheduler/issues/600

### DIFF
--- a/lib/resque/scheduler/lock/resilient.rb
+++ b/lib/resque/scheduler/lock/resilient.rb
@@ -43,7 +43,7 @@ module Resque
           @locked_sha = nil if refresh
 
           @locked_sha ||=
-            Resque.redis.script(:load, <<-EOF.gsub(/^ {14}/, ''))
+            Resque.data_store.redis.script(:load, <<-EOF.gsub(/^ {14}/, ''))
               if redis.call('GET', KEYS[1]) == ARGV[1]
               then
                 redis.call('EXPIRE', KEYS[1], #{timeout})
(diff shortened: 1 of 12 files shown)


lib/resque/scheduler/lock/resilient.rb
```diff
@@ -43,7 +43,7 @@ def locked_sha(refresh = false)
           @locked_sha = nil if refresh
 
           @locked_sha ||=
-            Resque.redis.script(:load, <<-EOF.gsub(/^ {14}/, ''))
+            Resque.redis.redis.script(:load, <<-EOF.gsub(/^ {14}/, ''))
```
jaredbeck: hmmm needs more redis' ;)

carsonreinke: I know.

```
2.3.3 :001 > Resque.redis.class
 => Resque::DataStore 
2.3.3 :002 > Resque.redis.redis.class
 => Redis 
```

Calling `Resque.redis.script` delegates to `Redis::Namespace`, `Resque.redis.redis.script` delegates to `Redis`.  Scripts do not need a namespace.

josephpage:
Method `Resque::DataStore#redis` relies on [`Resque::DataStore#method_missing`](http://www.rubydoc.info/gems/resque/Resque/DataStore#method_missing-instance_method)


So `Resque::DataStore#redis` returns the return of `Redis::Namespace#redis`. Maybe we should use `Redis::Namespace#_client` instead, as suggested in the code : https://github.com/resque/redis-namespace/blob/152253f5d38fcf325f288e881e7ade57c0e346bd/lib/redis/namespace.rb#L255

BTW, I am glad that this PR is merged in the near future.

carsonreinke:
Fixes #632 and @josephpage suggestion for another deprecation.

Issue #605, opened 2017-04-19
Support Duration for jobs periodicity


creends:
Instead of doing
```ruby
name = 'send_emails'
config = {}
config[:class] = 'SendEmail'
config[:args] = 'POC email subject'
config[:every] = ['1h', {first_in: 5.minutes}]
config[:persist] = true
Resque.set_schedule(name, config)
```

it would be a good idea to be able to do this
```ruby
config[:every] = [1.hour, {first_in: 5.minutes}]
```

carsonreinke:
The whole `.hour` and `.minutes` would require a dependency on [ActiveSupport](http://www.rubydoc.info/gems/activesupport/5.1.0/ActiveSupport/Duration).

Though, maybe we could just have it take a Numeric of seconds, because...

```
$> 1.hour.to_i
 => 3600 
$> 1.hour.is_a?(Numeric)
 => true 
```

PR #607, opened 2017-05-19
Allow duration to be used for scheduling every

carsonreinke:
Feature based on https://github.com/resque/resque-scheduler/issues/605.

Since rufus-scheduler allows for numeric values for `every`, durations can be used like `30.minutes` and `1.hour`.


test/scheduler_test.rb
```diff
@@ -132,6 +132,19 @@
...
+  test 'load_schedule_job with every fixnum' do
+    Resque::Scheduler.load_schedule_job(
+      'some_ivar_job',
+      'every' => 30,
+      'class' => 'SomeIvarJob',
+      'args' => '/tmp'
+    )
+
+    assert_equal(1, Resque::Scheduler.rufus_scheduler.jobs.size)
+    assert_equal(1, Resque::Scheduler.scheduled_jobs.size)
```
josephpage: Why not test the new job is well scheduled every 30s, and not 30m or 30ms ?

lib/resque/scheduler.rb
```diff
@@ -136,7 +136,7 @@ def load_schedule_job(name, config)
           interval_defined = false
           interval_types = %w(cron every)
           interval_types.each do |interval_type|
-            next unless !config[interval_type].nil? && !config[interval_type].empty?
+            next unless !config[interval_type].nil? && !config[interval_type].to_s.empty?
```
iloveitaly, line 139: What is this fixing?

test/scheduler_test.rb
iloveitaly: agreed, we need to test the actual schedule that's configured in the scheduler here.

iloveitaly (changes requested):
This is a great change! Left a couple comments, but we also need to:

* Update the changelog
* Rebase on master so CI runs properly

lib/resque/scheduler.rb
carsonreinke: @iloveitaly I believe this is since the interval type can now be `Numeric` and `empty?` is a `String` only.

test/scheduler_test.rb
carsonreinke: @josephpage @iloveitaly definitely, updated!

PatrickTulskie:
@carsonreinke hey just a heads up, I'm making moves to get through a bunch of the PR backlog for resque scheduler. This looks like a good candidate for 5.1.0. If you have a chance, can you rebase and push this up again and I'll get it live?

If you're not interested anymore or I don't hear back from you in the next week or so, I'll have my agent rebase and open a fresh PR with your change so you still get credit for the contribution.

Issue #597, opened 2017-02-07
Manual only schedule?


chrishough:
Is there a way to use scheduler for manual jobs?  I have been using rake for things like this but I was curious.

carsonreinke:
Kind of defeats the purpose of a scheduled job, @chrishough what were you thinking specifically?  Are you trying to use the "Queue now" portion of resque-web?

chrishough:
Yes it does to some extent.  I was thinking of handy way to incorporate running manual jobs that are not setup in cron via the same interface.  This way developers would not have to go in via ssh and rake.

carsonreinke:
@chrishough I ran a test and was able to do this:

Within a static schedule, I added this schedule item:

```
"Static Job":
  class: StaticJob
  never: "* * * * *"
```

The Rake task just complains with `no cron / every found for StaticJob (Static Job) - skipping` and it shows up in the resque-web interface!

![screenshot from 2017-02-13 14-03-20](https://cloud.githubusercontent.com/assets/378481/22898413/363a4aa2-f1f5-11e6-82a8-8f1a2b9b6d14.png)

:boom:

carsonreinke:
Go for it!!!

chrishough:
Thank you @carsonreinke 👍  Sorry it took me so long to put this together.

PR #608, opened 2017-05-21
features/update readme for manual jobs

chrishough:
Per the ticket https://github.com/resque/resque-scheduler/issues/597 add subsequent discussion with @carsonreinke I have updated the readme to show an example of using resque-scheduler for manual job execution.


chrishough:
Thank you @carsonreinke

Issue #484, opened 2015-07-29
Wait for enqueue success before recording last enqueued at


greggroth:
We just hit a case in our application where we ran the `resque-scheduler` process without loading Rails, so it didn't have access to our job classes to get the queue and the queue wasn't specified in the schedule YAML file.  Our resque-scheduler status page shows the "Last Enqueued" looked right, but the job never ran.  I suspect it's because the [`last_enqueued_at` is recorded before it's known whether the job was successfully enqueued](https://github.com/resque/resque-scheduler/blob/master/lib/resque/scheduler.rb#L141).

Is that intended behavior or can I open a PR to wait for the job to be enqueued before recording the timestamp?

carsonreinke:
Yeah, should be flip-flopped.  Throw a PR, if not I can.

PR #611, opened 2017-05-24
Wait for enqueue success before recording last enqueued at

carsonreinke:
Fixing https://github.com/resque/resque-scheduler/issues/484


kchudy3:
Any luck @carsonreinke @tmedford ?

Cryptophobia:
Is it time to think about merging this in and fixing the problem first identified in 2015? Or should we just keep praying. 🙏

Cryptophobia:
Thanks @carsonreinke !

PR #619, opened 2017-09-26
Travis tests with latest ruby engines

josephpage:
Resque-Scheduler should be tested with the same major Ruby versions than Resque itself.

Source : https://github.com/resque/resque/blob/master/.travis.yml

Thanks for reviewing !


fw42:
Thanks!

PR #620, opened 2017-09-29
redis-rb 4.0 support

jeremy:
* Loosen Redis version dep to `>= 3.3, < 5`
* Bump redis-namespace for looser Redis version dep.

No [API breakage](https://github.com/redis/redis-rb/blob/master/CHANGELOG.md#40) to deal with.


ota42y:
Hi! redis-namsepace 1.6.0 released!
So we don't need set github repository to Gemfile

jeremy:
Gentle nudge. Rails 5.2 beta is coming soon and it expects Redis 4.0. Let's get a new gem out. (I'm happy to help with ongoing maintenance and releases.) /cc @fw42 @carsonreinke @hkdsun @meatballhat

carsonreinke:
Added #626, I would add you if I could.

PR #627, opened 2017-11-27
Redis timing out will no longer crash the whole process.

crazymykl:
This seems to occur most often when updating dynamic schedules, but it can happen other times.

A good way to test that this fixes something is to run the scheduler, send redis-server SIGTSTP, and see if the scheduler blows up the next time it needs to talk to redis.


lib/resque/scheduler.rb
```diff
@@ -13,6 +13,9 @@ module Scheduler
     autoload :Extension, 'resque/scheduler/extension'
     autoload :Util, 'resque/scheduler/util'
     autoload :VERSION, 'resque/scheduler/version'
+    INTERMITTENT_ERRORS = [
```
crazymykl: An early version of this PR called these `CRASHY_BOIS`.

crazymykl:
This has decreased the amount of alerts we get in production, so it seems to work correctly.

carsonreinke:
@crazymykl I tried your change, but I am only receiving `Connection timed out` instead of the process terminating.  Also, killing Redis keeps the process open too. Are you coming across this in the real world?  Have you tried this change?

I tried the Resque worker process to see what happened and that definitely terminates.

I feel like any change here would would require a major bump.

crazymykl:
> I tried your change, but I am only receiving Connection timed out instead of the process terminating.

That's the intent of this patch. Redis flapping or going down won't end the scheduler process. 

My company is running this in production, since we were having issues with scheduler processes vanishing due to AWS Elasticache occasionally taking too long to respond.

crazymykl:
It was specifically redis timeouts during the update of a dynamic schedule bringing the process down.

carsonreinke:
@crazymykl Oh I see, this is to prevent a crash from this.

carsonreinke:
Merged in #636

Issue #633, opened 2018-01-16
scheduler can NOT enqueue job correctly sometimes


ceshihao:
I use `resque-scheduler` v4.0.0

Recently, my resque schedule can NOT enqueue job correctly sometimes.
There are 3 phenomenons, but i can NOT find root cause so far.
1. Sometimes, all scheduled jobs can NOT be enqueued automatically, but they can be triggered from resque web.
2. Sometimes, some of jobs can be enqueued automatically, but some are not.
3. Sometimes, all jobs work correctly.

I have some hints but not sure.
* `dynamic schedule` is enabled.
* I have 2 scheduler for HA.

carsonreinke:
@ceshihao Well that is concerning.

In the resque-web, do you see all the scheduled jobs?

ceshihao:
I found it is because of an incompatible problem between `static` and `dynamic` schedules.

- Previously, I use static scheduler and load the schedules from a `yaml`. It works and No schdule lost.

- Recently, I try to use `dynamic` one. However, `dynamic` scheduler thinks the schedules are temporary ones if NOT set `persist` = `true`. Therefore, when one scheduler is killed for some reason, original static ones are cleared (v4.0.0 behavior).

carsonreinke:
@ceshihao It seems like changing the default to `persist: true` would make the most sense and prevent the issue you had in the future?

PR #637, opened 2018-02-02
change dynamic schedule set_schedule default persist

ceshihao:
To improve the incompatible problem #633


carsonreinke:
@ceshihao Totally agree with this.  I believe this would be a breaking change with semantic versioning, so it would have to be included in `v5`.

There is some test failures from the default change, probably just simply need to set `persist: false` for this cases.

PR #639, opened 2018-02-26
Lua zpop

xldenis:


rafaelfranca:
We are using a fork on resque-scheduler because we made some backward incompatible changes. I'm guessing this PR was to our fork.

I think we should upstream everything we have in our fork even it is backward incompatible and release a major version. Maybe I'll do that some time in the future.

The PR to our fork is here https://github.com/Shopify/resque-scheduler/pull/23.

jeremy:
> upstream everything we have in our fork even it is backward incompatible and release a major version

Yes, please! 😮👉 https://github.com/Shopify/resque-scheduler/pulls?utf8=✓&q=is%3Aclosed+

xldenis:
Yea, the idea is to eventually port this to upstream but we need to figure out a satisfactory migration plan for everyone to move from the totally incompatible previous backend.

PR #661, opened 2018-11-24
Support resque 2.x.x

pieterza:
- Test suite passes locally.  Failed with travis here due to missing REDIS env, not sure why.
- Major version bump upstream resque mainly to drop support for older ruby/rails.  This is okay as resque-scheduler will still be able to use lower versions than resque 2.0.0 provided users have it locked as such.

### DIFF
--- a/resque-scheduler.gemspec
+++ b/resque-scheduler.gemspec
@@ -55,6 +55,6 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'mono_logger', '~> 1.0'
   spec.add_runtime_dependency 'redis', '>= 3.3'
-  spec.add_runtime_dependency 'resque', '>= 1.26', '< 3.0.0'
+  spec.add_runtime_dependency 'resque', '>= 1.26'
   spec.add_runtime_dependency 'rufus-scheduler', '~> 3.2'
 end
--- a/resque-scheduler.gemspec
+++ b/resque-scheduler.gemspec
@@ -55,6 +55,6 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'mono_logger', '~> 1.0'
   spec.add_runtime_dependency 'redis', '>= 3.3'
-  spec.add_runtime_dependency 'resque', '>= 1.26', '<= 2.0.0'
+  spec.add_runtime_dependency 'resque', '>= 1.26', '< 3.0.0'
   spec.add_runtime_dependency 'rufus-scheduler', '~> 3.2'
 end
--- a/resque-scheduler.gemspec
+++ b/resque-scheduler.gemspec
@@ -55,6 +55,6 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'mono_logger', '~> 1.0'
   spec.add_runtime_dependency 'redis', '>= 3.3'
-  spec.add_runtime_dependency 'resque', '~> 1.26'
+  spec.add_runtime_dependency 'resque', '>= 1.26', '<= 2.0.0'
   spec.add_runtime_dependency 'rufus-scheduler', '~> 3.2'
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,11 @@ matrix:
   - rvm: jruby-9.1.13.0
 services:
 - redis-server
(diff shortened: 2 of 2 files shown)


resque-scheduler.gemspec
```diff
@@ -55,6 +55,6 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'mono_logger', '~> 1.0'
   spec.add_runtime_dependency 'redis', '>= 3.3'
-  spec.add_runtime_dependency 'resque', '~> 1.26'
+  spec.add_runtime_dependency 'resque', '>= 1.26', '<= 2.0.0'
```
jeremy: Note that this supports 2.0.0 but not 2.0.1 or later

rafaelfranca:
```suggestion
  spec.add_runtime_dependency 'resque', '>= 1.26'
```

pieterza: Correct, I can only prove that 2.0.0 works at this time.

pieterza: Major version bumps are typically known to break backwards compatibility, but in this case 2.0.0 does not.  It's possible for future 2.0.1 to actually be a breaking change, so I erred on the side of caution.

pieterza: >= 1.26 would also allow future 3.0.0 which is most likely not what we want.

pieterza: To be fair, by my own logic a major version bump typically contains breaking changes.
Thus since 2.0.0 did not break resque-scheduler, 2.x.x should not either.

In that case, we should change it to `spec.add_runtime_dependency 'resque', '>= 1.26', '< 3'`
`

pieterza: To be fair, by my own logic a major version bump typically contains breaking changes.
Thus since 2.0.0 did not break resque-scheduler, 2.x.x should not either.

In that case, we should change it to `spec.add_runtime_dependency` 'resque', '>= 1.26', '< 3'`

pieterza: Ahh, okay.  Apologies for resolving conversation.  I've changed as per request.

pieterza: I've changed as per request based on reply by @rafaelfranca

rafaelfranca: No need to apologize. Your logic makes sense and it is what most projects do. We were being doing that for years in Rails but we recently changed since no release is a breaking change with the previous release.

pieterza: Thank you!  Resolving then.

rafaelfranca:
CI is broken and we can't merge. Can you take a look?

tmedford:
We could really use this to be pulled in. Thanks.

pieterza:
Bumpity bump .

rafaelfranca:
CI is still broken and we can't merge. Can you take a look?

tmedford:
Please take a look

jeremy:
Nice CI fix. Thanks!

pieterza:
FYI, this was fixed by disabling potential IPv6, and ensuring redis responds as a kind of health check.
I assume the upstream redis-server travis package was changed since last build.

Fix:
`
before_install:
  - sudo sed -e 's/^bind.*/bind 127.0.0.1/' /etc/redis/redis.conf > redis.conf
  - sudo mv redis.conf /etc/redis
  - sudo service redis-server start
  - echo PING | nc localhost 6379`

Credit to: https://travis-ci.community/t/trusty-environment-redis-server-not-starting-with-redis-tools-installed/650/2

ygnessin:
Thanks for this! Is there any ETA on when a new gem version will be released containing this change?

Techn1x:
I'd really like to upgrade to Redis 4 which requires me to update to Resque 2 (adds support for Redis 4) which means I need resque-scheduler to add support for Resque 2... any chance we can cut a release?

rafaelfranca:
You can always use the version from github since bundler is our friend and allow that. A release will be cut when the maintainer think it is ready and when they have time.

jeremywadsack:
As far as I know, you can't use a version from github when you are setting requirements in a gemspec. A released version would allow us to update gems that depend on `resque-scheduler`.

Issue #677, opened 2019-06-19
removing many dynamic schedules at once is not performant


ledhed2222:
I have a situation where ~80K dynamic jobs were created before we realized that this approach was not going to work (we're using a static schedule now that is much saner). To clean up all of the persisted schedules using the public interface is straightforward:

```ruby
(Resque.schedule.keys - YAML.load_file("path", "to", "schedule", "file").keys).each do |to_remove|
  Resque.remove_schedule(to_remove)
end
```

However, this will not only make one `hdel` call to redis for each key, but will also re-load the entire schedule each time!

I can imagine a simple change to the `remove_schedule` method that would allow a variable number of schedules to be removed at once:

```ruby
def remove_schedule(*schedules)
  schedules.each { |schedule| non_persistent_schedules.delete(schedule) }
  redis.hdel(:persistent_schedules, *schedules)
  redis.sadd(:schedules_changed, *schedules)
  reload_schedule!
end
```

What do you guys think about that? Happy to make a PR if that makes sense.

ledhed2222:
actually - it appears that we'd only have to update the `schedules_changed` set with one entry here. updated my proposal above.

ledhed2222:
Perhaps a different suggestion. It seems like the root issue here is that there is no story to migrate _from_ dynamic schedules back to a static schedule, if you make the same mistake we did. Instead of the above idea where you can remove many schedules at once, what about purging the entire dynamic schedule:

```ruby
def purge_dynamic_schedule!
  if !dynamic
    redis.del(:persistent_schedules)
    redis.del(:schedules_changed)
    reload_schedule!
  end
end
```

this way it's "opt-in": users may wish to drain their dynamic schedules naturally after moving off. but they could optionally just kill the entire

jeremy:
@ledhed2222 Bulk removal is welcome! Purging as well. Good call.

ledhed2222:
howdy. i'm working on this PR but I'm having trouble passing the test suite. could you provide some insight?

```
Error: test_concurrent_shutdowns_and_startups_do_not_corrupt_the_schedule(Multi_Process): RuntimeError: forked process failed with pid 43772 SIGABRT (signal 6)
/Users/gregweisbrod/Documents/Dev/resque-scheduler/test/multi_process_test.rb:111:in `block in get_results_from_children'
/Users/gregweisbrod/Documents/Dev/resque-scheduler/test/multi_process_test.rb:108:in `each'
[...]

PR #686, opened 2019-12-15
add #purge_dynamic_schedule! method

ledhed2222:
see https://github.com/resque/resque-scheduler/issues/677

### DIFF
--- a/test/scheduler_test.rb
+++ b/test/scheduler_test.rb
@@ -431,7 +431,7 @@ context 'Resque::Scheduler' do
     Resque.remove_schedule('some_ivar_job3')
     assert_equal nil, Resque.redis.hget(:schedules, 'some_ivar_job3')
     assert Resque.redis.sismember(:schedules_changed, 'some_ivar_job3')
-    assert_equal [], Resque.redis.smembers(:persisted_schedules)
+    assert_equal [], Resque.redis.smembers(:persistent_schedules)
   end
 
   test 'remove_schedule does not reload schedule when disabling reload flag' do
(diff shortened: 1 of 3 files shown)


test/scheduler_test.rb
```diff
@@ -431,7 +431,7 @@
     Resque.remove_schedule('some_ivar_job3')
     assert_equal nil, Resque.redis.hget(:schedules, 'some_ivar_job3')
     assert Resque.redis.sismember(:schedules_changed, 'some_ivar_job3')
-    assert_equal [], Resque.redis.smembers(:persisted_schedules)
+    assert_equal [], Resque.redis.smembers(:persistent_schedules)
```
ledhed2222, line 434: I think this was a typo previously

test/scheduler_test.rb
```diff
@@ -467,10 +467,10 @@
     assert_equal true, Resque.schedule['some_ivar_job5'].nil?
     assert Resque.redis.sismember(:schedules_changed, 'some_ivar_job4')
     assert Resque.redis.sismember(:schedules_changed, 'some_ivar_job5')
-    assert_equal [], Resque.redis.smembers(:persisted_schedules)
+    assert_equal [], Resque.redis.smembers(:persistent_schedules)
```
ledhed2222, line 470: Ditto

test/scheduler_test.rb
```diff
@@ -585,6 +585,44 @@ def enqueue_started!
...
+        'some_ivar_job',
+        'cron' => '* * * * *', 'class' => 'SomeIvarJob', 'args' => '/tmp/2 ',
+        'persist' => true
+      )
+      assert_equal 1, Resque.redis.scard(:schedules_changed)
+      assert_equal 1, Resque.redis.hlen(:persistent_schedules)
+      Resque.purge_dynamic_schedule!
+      assert_equal 0, Resque.redis.scard(:schedules_changed)
+      assert_equal 0, Resque.redis.hlen(:persistent_schedules)
+    end
```
ledhed2222, line 623: note, not testing that `reload_schedule!` is called, because it seems difficult to mock all of that out and the core part here is that redis keys are being deleted. lmk what you think.

ledhed2222:
@jeremy finally got around to this. looks to me like the failing test also is failing on `master` and isn't my addition :(

i've added some comments in the pr on a few things of note

PR #688, opened 2019-12-25
Fix XSS vulnerability

inatomitk:
When I did vulnerability diagnosis at my company, the vulnerability scanner found the delayed/search page is susceptible to XSS.
This fixes that issue.


pradyumna2905:
One of our client's diagnostic tool is also reporting the same vulnerability.
Some more info on that here: https://cwe.mitre.org/data/definitions/79.html

hathawad:
Can we get a release with this change in it?

PR #689, opened 2019-12-27
Ensure that jobs are directly created for jobs enqueued at Time.now

jdlubrano:
# Problem

If less than a second passes between calling `Resque.enqueue_in(0, ...)`, which is not all that unlikely, a job may be delayed when we don't expect the job to be delayed.  In high traffic applications where the Resque Scheduler process is very busy, this unexpected delay can grow as other jobs are being enqueued after their own delays.  Since Resque Scheduler only supports multiple instances for redundancy, there is no way to scale Resque Scheduler horizontally.  Thus my goal is to ensure that as many jobs as possible are created directly and not unexpectedly delayed.

# Solution

Use a `<=` comparison with `Time.now` to determine whether or not to delay enqueuing/creating a Resque job.  For what it's worth, this PR is reminiscent of https://github.com/resque/resque-scheduler/pull/226, but I want to use `<=` rather than `<` for the `Time.now` comparison.

### DIFF
--- a/lib/resque/scheduler/delaying_extensions.rb
+++ b/lib/resque/scheduler/delaying_extensions.rb
@@ -24,7 +24,7 @@ module Resque
       def enqueue_at_with_queue(queue, timestamp, klass, *args)
         return false unless plugin.run_before_schedule_hooks(klass, *args)
 
-        if Resque.inline? || timestamp.to_i < Time.now.to_i
+        if Resque.inline? || timestamp.to_i <= Time.now.to_i
           # Just create the job and let resque perform it right away with
           # inline.  If the class is a custom job class, call self#scheduled
           # on it. This allows you to do things like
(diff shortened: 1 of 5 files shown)


jdlubrano:
@jeremy Sorry for the direct ping, but any thoughts about this?  I am happy to provide any additional tests or context regarding the problem that I am trying to solve if necessary.  Thanks!

jdlubrano:
@jeremy 

Thanks for accepting this PR!

How are releases of this gem usually handled?  Selfishly I would like to use this change in production by simply `bundle` updating the `resque-scheduler` gem.  I could tweak my Gemfiles to refer to the master branch, but I think it would be great to see this change released in a patch or minor version bump.  I am not sure if this PR changes enough functionality to warrant a minor release or not either...hopefully the behavior introduced here is desirable to all (or mostly all) users of this gem 😀.

jeremy:
Cheers! Single PRs don't trigger a release, however. This'll go out with the next minor version bump.

PR #694, opened 2020-04-08
add a railtie to load rescue scheduler automatically in rails apps

josh-m-sharpe:
Should eliminate some basic boilerplating.


lib/resque/scheduler.rb
```diff
@@ -8,6 +8,8 @@
 require_relative 'scheduler/signal_handling'
 require_relative 'scheduler/failure_handler'
 
+require 'resque/scheduler/railtie' if defined?(Rails)
```
iloveitaly, line 11: Based on the issue that came up in `resque` I think we need to do

```suggestion
require 'resque/scheduler/railtie' if defined?(Rails::Railtie)
```

Right?

iloveitaly (changes requested):
Thanks for this! Could you:

* Rebase on master so CI runs
* Add a note to the changelog
* Update the readme?

Thanks!

PatrickTulskie:
@josh-m-sharpe hey just a heads up, I'm making moves to get through a bunch of the PR backlog for resque scheduler. This looks like a good candidate for 5.1.0. If you have a chance, can you rebase and push this up again and I'll get it live? Also, it looks like the changes that @iloveitaly will need to be done for this to get merged.

If you're not interested anymore or I don't hear back from you in the next week or so, I'll have my agent rebase and open a fresh PR with your change so you still get credit for the contribution.

PR #696, opened 2020-04-08
add support for queueing ActiveJob jobs

josh-m-sharpe:
This allows resque scheduler to set the queue correctly for ActiveJobs when resque schedule's YAML configuration does not set the queue.  Depends on this resque PR:  https://github.com/resque/resque/pull/1713

### DIFF
--- a/lib/resque/scheduler.rb
+++ b/lib/resque/scheduler.rb
@@ -276,6 +276,8 @@ module Resque
             # CustomJobClass). Otherwise, pass off to Resque.
             if klass.respond_to?(:scheduled)
               klass.scheduled(queue, klass_name, *params)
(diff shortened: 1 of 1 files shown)


lib/resque/scheduler.rb
```diff
@@ -276,6 +276,8 @@ def enqueue_from_config(job_config)
             # CustomJobClass). Otherwise, pass off to Resque.
             if klass.respond_to?(:scheduled)
               klass.scheduled(queue, klass_name, *params)
+            elsif klass.respond_to?(:perform_later)
+              klass.set(queue: queue).perform_later(*params)
```
iloveitaly, line 280: Could you add a one-line comment here about what is code is doing? Even just linking to this PR would be helpful for the future reader.

PR #698, opened 2020-06-24
Refactor remove_delayed_selection to allow convenient massive jobs cleanup

kapcod:
The need for this change arised when we had incident that caused scheduling of 700,000 invalid jobs that had to be cleaned.
This method deletes the jobs as it finds them, so if the process is interrupted it can be resumed.

Here's why I didn't use the existing methods:
`find_delayed_selection` - would load all the matching jobs in memory, high memory usage and impossible to track progress of the scanning
`remove_delayed_job` - is inconvenient as it tries to delete the job from all timestamps, which is not needed here, because it goes over all timestamps anyway
`remove_delayed_job_from_timestamp` - seemed to do what's needed but it expects decoded arguments and reencodes them, which is unnecessary computation and it calls `clean_up_timestamp`, which previously wasn't done as part of this method.

### DIFF
--- a/lib/resque/scheduler/delaying_extensions.rb
+++ b/lib/resque/scheduler/delaying_extensions.rb
@@ -174,13 +174,33 @@ module Resque
       #
       # This allows for removal of delayed jobs that have arguments matching
       # certain criteria
-      def remove_delayed_selection(klass = nil)
+      # progress - if set to true will output progress to STDOUT
+      def remove_delayed_selection(klass = nil, progress: false, &block)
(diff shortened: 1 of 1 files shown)


kapcod:
The search itself got much slower in a system with thousands of different delayed timestamps. This implementation is good when need to remove a lot of jobs, but actually much slower when need to find and remove just a few jobs. Closed the pull request in favor of just keeping this code aside and running it manually.
It can also be added as an additional method in this repository.

PR #701, opened 2020-11-12
Delaying & Scheduling extension: Wrap ops in multi/exec

pjambet:
Wrap write operations in multi/exec blocks to prevent any possibilities
of Redis being in an inconsistent state. It also has the added benefit
of providing small optimizations by reducing the number of round-trips.

The commit uses a verbose approach of wrapping most calls to
`redis.multi` in a `redis.pipelined` block, which is _technically_
unnecessary since the redis-rb gem does not send the `multi` & `exec`
command if it is pipelining the commands.

The reason this commit uses both is that both methods have different
semantics, `pipelined` is meant to be an optimization whereas `multi`
provides the guarantees of a Redis transaction.

While it may seem unlikely, it seems possible that future versions of
the redis gem change the underlying behavior of the `multi` and
`pipelined` commands. Additionally, it is a little bit more explicit in
terms of describing intent: "I want these commands to be run in a atomic
fashion, and I'm ok sending it all at once".

The call to `redis.del(key)` in `clean_up_timestamp` was unnecessary
since the only reason that would cause the `llen` call above to return
`0` is if the list did not exist.

The call to `pipelined` in this example might seem even more overkill
since we only give a single command to `multi`, but `multi`/`exec` are
themselves commands, so in the eventuality that the redis gem would
start sending the `multi` command right away in a future version,
wrapping it in a `pipelined` call is explicitly asking it to send
`multi`/`zrem`/`exec` all at once.

---

I totally understand that the `redis.pipelined/redis.multi` approach I took in this commit is kinda pedantic and given that a call to `pipelined` is enough to guarantee atomicity, we could drop a bunch of `multi` calls in this commit and maintain the same behavior, with the exception of the existing `multi` call in `clean_up_timestamp` since none of them use `WATCH` or `DISCARD`.

In there we _could_ drop the `pipelined` call, and still benefit from the ruby client pipelining the commands anyway, but that would be less explicit. Happy to hear what other think

### DIFF
--- a/lib/resque/scheduler/delaying_extensions.rb
+++ b/lib/resque/scheduler/delaying_extensions.rb
@@ -68,16 +68,20 @@ module Resque
       # insertion time complexity is O(log(n)). Returns true if it's
       # the first job to be scheduled at that time, else false.
       def delayed_push(timestamp, item)
-        # First add this item to the list for this timestamp
-        redis.rpush("delayed:#{timestamp.to_i}", encode(item))
+        redis.pipelined do
+          redis.multi do
+            # First add this item to the list for this timestamp
+            redis.rpush("delayed:#{timestamp.to_i}", encode(item))
 
-        # Store the timestamps at with this item occurs
-        redis.sadd("timestamps:#{encode(item)}", "delayed:#{timestamp.to_i}")
+            # Store the timestamps at with this item occurs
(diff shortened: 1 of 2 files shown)


pjambet:
Tests failed for ruby 2.6 and resque ~1.27, trying to replicate locally, need to install the right ruby version and all that

PR #709, opened 2021-04-27
fix URI.escape and URI.encode

YoogoC:
ruby 3 removed URI.escape and URI.encode~


iloveitaly:
@and9000 @bubba going to merge this in; would one of you have the time to write another PR to correct the other instances of this issue? Looks like some of the work was already done here https://github.com/resque/resque-scheduler/pull/710/files

PR #711, opened 2021-06-01
rufus-scheduler picks timezone from rails configuration

light94:
Since rufus-schedule 3.3.3 time zone is taken from rails config if set. https://github.com/jmettraux/rufus-scheduler/blob/64591be6a2c80149839861389a0ed8708e279198/CHANGELOG.md#rufus-scheduler-333---released-2017-01-29

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,9 +407,6 @@ Resque.set_schedule(name, config)
 
 #### Time zones
 
-Note that if you use the cron syntax, this will be interpreted as in the server time zone
-rather than the `config.time_zone` specified in Rails.
-
 You can explicitly specify the time zone that rufus-scheduler will use:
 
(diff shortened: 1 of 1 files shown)


shuuuuun:
I was also misled by this document and used it.
I hope it will be merged.

light94:
@iloveitaly this is a minor change in the Readme which will help reduce confusion for future users. Please do look at it when you get a chance

iloveitaly:
@light94 doesn't this still apply if the user doesn't have rails? Could we adjust the language and call out the existing functionality still holds if rails is not in place?

light94:
@iloveitaly I've tried to incorporate your remarks, does it look better to you?

PR #715, opened 2021-07-11
Allow jobs to declare theirs schedule

and9000:
Hello,

I've implemented the possibility for the job to declare it's schedule so it's self contained.


and9000:
@jeremy sorry for pinging you directly, do you have any though about this? Thanks!

Btw as we are here 😄 I work for a company ([Nozomi Networks](https://www.nozominetworks.com/) - OT & IoT Security
and Visibility) which would like to contribute and maintain this project (and moving on to resque itselt) as we heavily use ruby and rely on resque.

Thanks!

iloveitaly:
@and9000 Would love your help maintaining this!

The first thing we'll need to do is get CI working again. Could you submit a separate PR to move CI to GH actions? Here's what we did for the core resque gem: https://github.com/resque/resque/blob/master/.github/workflows/ci.yml

and9000:
Sure, I'll start working on CI 👍

lib/resque/scheduler/job.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        def every(value = nil)
+          return @every ||= nil if value.nil?
+          @every = value
+        end
+
+        def queue(value = nil)
+          return @queue ||= nil if value.nil?
+          @queue = value
+        end
```
iloveitaly, line 26: The standard Resque convention is to use `@queue` without a class method. I think we should continue to use that convention if we can.

lib/resque/scheduler.rb
```diff
@@ -107,6 +107,26 @@ def load_schedule!
...
+
+        Dir[auto_load.to_s].each do |file|
+          require File.absolute_path(file)
+          name = File.basename(file, '.rb')
+          begin
+            klass = Resque::Scheduler::Util.constantize(name)
+          rescue NameError
+            log! "Can't load file #{file}"
+          end
+          load_schedule_job(
```
iloveitaly, line 119: Is it possible to run `load_schedule_job` when the schedule is set up on the class? If we used a `resque_schedule` method on the class-level as suggested below, do you think we could eliminate the auto-load logic from the PR and instead run `load_schedule_job` as part of `resque_schedule`.

Autoloading and scheduling on the class-level feel like two different concerns to me, and it's not clear why we should couple them in this case.

Additionally, autoloading feels like something that makes sense on the core resque level as opposed to the scheduler plugin.

lib/resque/scheduler/job.rb
```diff
@@ -0,0 +1,40 @@
...
+  module Scheduler
+    module Job
+      class << self
+        def included(base)
+          base.extend ClassMethods
+        end
+      end
+
+      module ClassMethods
+        def cron(value = nil)
```
iloveitaly, line 13: If feels like it could be cleaner to have a single class method `resque_schedule` which has a bunch of optional args (`cron:`, `every:`, etc). I don't love the idea of adding a bunch of methods that aren't clearly ties to resque scheduler (i.e. `description` could easily be a method users of this gem have defined on their class).

What do you think?

iloveitaly (changes requested):
Awesome idea! I love the idea of (optionally) grouping the scheduling logic with the class definition—makes it much more clear to the reader whats going on.

Couple thoughts about the specific implementation, but I'd love to get this change in.

lib/resque/scheduler/job.rb
and9000: You are right. Fixed.

lib/resque/scheduler/job.rb
and9000: Yes code looks cleaner this way 😄

lib/resque/scheduler.rb
and9000: I've moved schedule logic into class method `resque_schedule`.

I'm still stuck thinking on where the autoloading feature should be: in principle yes it shouldn't be here but on the application level (not even resque) as the app is responsible to load stuff. But this mean we should add a big warning "in order to self-schedule-jobs to work please remember to require your jobs". But this is exactly what `auto_load` is doing so my doubt is: why don't make that requirement explicit?

PR #706, opened 2021-03-15
Require rufus-scheduler < 3.7

hosamaly:
With v3.7, we see the following error from `multi_process_test`:

```
resque-scheduler/vendor/bundle/ruby/2.4.0/gems/rufus-scheduler-3.7.0/lib/rufus/scheduler.rb:598:in `pop': No live threads left. Deadlock? (fatal)
1 threads, 1 sleeps current:0x00000000017765e0 main thread:0x00000000017765e0
* #<Thread:0x00000000017a9e18 sleep_forever>
   rb_thread_t:0x00000000017765e0 native:0x00007f454bb51700 int:0
   resque-scheduler/vendor/bundle/ruby/2.4.0/gems/rufus-scheduler-3.7.0/lib/rufus/scheduler.rb:598:in `pop'
   resque-scheduler/vendor/bundle/ruby/2.4.0/gems/rufus-scheduler-3.7.0/lib/rufus/scheduler.rb:598:in `no_time_limit_join'
   resque-scheduler/vendor/bundle/ruby/2.4.0/gems/rufus-scheduler-3.7.0/lib/rufus/scheduler.rb:148:in `join'
   resque-scheduler/lib/resque/scheduler.rb:380:in `stop_rufus_scheduler'
   resque-scheduler/lib/resque/scheduler.rb:384:in `before_shutdown'
   resque-scheduler/test/multi_process_test.rb:70:in `block (4 levels) in <top (required)>'
   resque-scheduler/test/multi_process_test.rb:92:in `block in fork_with_marshalled_pipe_and_result'
   resque-scheduler/test/multi_process_test.rb:89:in `fork'
   resque-scheduler/test/multi_process_test.rb:89:in `fork_with_marshalled_pipe_and_result'
   resque-scheduler/test/multi_process_test.rb:64:in `block (3 levels) in <top (required)>'
   resque-scheduler/test/multi_process_test.rb:63:in `times'
   resque-scheduler/test/multi_process_test.rb:63:in `block (2 levels) in <top (required)>'

E
================================================================================
Error: test_concurrent_shutdowns_and_startups_do_not_corrupt_the_schedule(Multi_Process): RuntimeError: forked process failed with pid 7680 exit 1
resque-scheduler/test/multi_process_test.rb:111:in `block in get_results_from_children'
resque-scheduler/test/multi_process_test.rb:108:in `each'
resque-scheduler/test/multi_process_test.rb:108:in `get_results_from_children'
[...]

PR #716, opened 2021-09-14
GitHub Actions CI Workflow

and9000:
Hello,

this PR adds GitHub Actions CI Workflow and removes old travis workflow. Some open points:

* all builds will fail as we need this merged https://github.com/resque/resque-scheduler/pull/706 
* builds with ruby >= 3 will fail as `URI.encode/URI.escape` has been removed. There's a PR opened for this but it's not fixing every occurrence https://github.com/resque/resque-scheduler/pull/709
* rubygems integration: there's an [action](https://github.com/marketplace/actions/publish-to-rubygems) for that but it needs a `GITHUB_TOKEN` to be able to create a tag and a `RUBYGEMS_API_KEY` to be able to push (they should be added as secrets to the repo). Obviously I don't have them 😄 

Thanks,
Andrea


iloveitaly:
Thanks so much for doing this!

PR #717, opened 2021-09-14
Setup a github actions workflow

bubba:


bubba:
You can see a sample run here: https://github.com/bubba/resque-scheduler/runs/3604291536?check_suite_focus=true

There seems to be one test failing with an exception being thrown from rufus-scheduler, but it looks like a legit failure, possibly due to a newer release of it

bubba:
@iloveitaly

iloveitaly:
@bubba thanks for this! Looks like someone else got to it first, going to close this out:

https://github.com/resque/resque-scheduler/pull/716

PR #718, opened 2021-09-15
Ruby 3 support

and9000:
* removed `URI.decode/URI.escape` in favour of `CGI.unescape/CGI.escape`
* using CGI.escape(args.to_json) instead of URI.encode_www_form(args)


iloveitaly:
Great, thanks!

Issue #712, opened 2021-06-11
Dalayed Job page is not loading in. rails 6 and ruby 3


shodhangk:
Hi,

When I tried to view the list of delayed jobs in the UI it is failing with the below error.

```
 NoMethodError - undefined method `escape' for URI:Module:
	/Users/shodhangk/.rbenv/versions/3.0.1/lib/ruby/gems/3.0.0/gems/resque-scheduler-4.4.0/lib/resque/scheduler/server.rb:132:in 
```

In ruby 3, the `escape` method is removed for URI. So alternative method need to be used.

YoogoC:
#709

sharang-d:
Could someone release a new version, please?

iloveitaly:
@sharang-d there's a couple things we need to do before a new release. Here's a list of you want to chip in with PRs:

https://github.com/resque/resque-scheduler/pull/671#issuecomment-924974183

iloveitaly:
This has been fixed https://github.com/resque/resque-scheduler/releases/tag/v4.5.0

PR #721, opened 2021-09-24
Remove Ruby 2.3 from testing matrix

sharang-d:
Mimic the setup in https://github.com/resque/resque/blob/master/.github/workflows/ci.yml

PR #722, opened 2021-09-24
Separate Rubocop and specs into two GH Action workflows

sharang-d:
What?
-
Fix CI!
As requested in https://github.com/resque/resque-scheduler/issues/712#issuecomment-925380319 :)

How?
-
Separate Rubocop and Ruby specs into two different GitHub Action workflows. 

Why?
-
Some builds were failing because Rubocop was failing on some Rubies. We don't need to run Rubocop against all Rubies so we now run Rubocop with only one ruby implementation.

I referred to https://github.com/rails/webpacker/tree/d7adcd20ddb77097a99a27f94491e78884c6ec59/.github/workflows for inspiration.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -1,0 +1,27 @@
+name: Rubocop
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  rubocop:
+    name: Rubocop
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        ruby: [
+            2.4
+        ]
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
+      - name: Ruby linter
+        run: bundle exec rubocop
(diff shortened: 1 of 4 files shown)


sharang-d:
@iloveitaly I think you (/someone else with access) needs to turn off the TravisCI required check.

sharang-d:
Do we still need AppVeyor? I don't think so. I will remove it after you confirm.

.github/workflows/rubocop.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  rubocop:
+    name: Rubocop
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        ruby: [
+            2.4
```
sharang-d, line 17: I used 2.4 here as I think [we don't need to support 2.3](https://github.com/resque/resque-scheduler/pull/721) but happy to change it if not.

iloveitaly: Any reason we shouldn't use the latest ruby version here instead?

iloveitaly:
This looks great, thank you!

.github/workflows/rubocop.yml
sharang-d: I think not. I went with the min supported instead of the max supported.

PR #729, opened 2021-11-03
Replace `Resque.redis.exists` with `Resque.redis.exists?`

mishina2228:
`Redis#exists` returns an Integer from [v4.5.0 of redis-rb](https://github.com/redis/redis-rb/blob/master/CHANGELOG.md#450).

This PR fixes the test failures as shown in the following link: 
https://github.com/resque/resque-scheduler/actions/runs/1417181019


iloveitaly:
Thank you!

PR #733, opened 2021-11-07
Fix view styles and typos

mishina2228:


iloveitaly:
Wow, great bunch of changes! Thank you.

PR #734, opened 2021-11-12
Use SET NX/EX for master locking instead of Lua script

### LINKED ISSUE #553: Master lock getting out of sync in case the LUA script gets interrupted
We had this production problem yesterday where we discovered that the scheduled tasks are not executed even though the scheduler is running. After investigation it turned out that the master lock key in redis is set to some value, but has no TTL set, essentially leading us to this function: https://github.com/resque/resque-scheduler/blob/master/lib/resque/scheduler/lock/resilient.rb#L54

The above inconsistency caused no master node to be elected (although we don't use multiple schedulers) and all the scheduled jobs got blocked.

I really believe the way this lock is set with 2 separate operations SETNX and EXPIRE is not atomic, even though it is executed in a LUA script. These 2 operations need to be atomic and this can be achieved using the `SET NX PX`. Even a better solution will be to use a lock implementation which is reviewed by the community, for example using the Redis guidelines for distributed locks: http://redis.io/topics/distlock

carsonreinke:
Fixes #553.  In general, a better and more accepted approach to the master locking.  Redis documentation even states:

> The command SET resource-name anystring NX EX max-lock-time is a simple way to implement a locking system with Redis.

https://redis.io/commands/set

**NOTE:** This does change Redis support slightly, but `resque-scheduler` will continue work.

Thanks @valo


lib/resque/scheduler/locking.rb
```diff
@@ -59,7 +60,11 @@ def master_lock
       end
 
       def supports_lua?
-        redis_master_version >= 2.5
+        redis_master_version >= Gem::Version.new('2.5')
+      end
+
+      def supports_resilient_lock?
+        redis_master_version >= Gem::Version.new('2.6.2')
```
yaauie, line 67: It looks like the extended `SET` options were available [since redis 2.6.12](https://github.com/redis/redis/issues/931#issuecomment-15764821).

yaauie (commented):
I'm all for simplifying calls where appropriate.

In this case though, any user who previously qualified for the resilient lock (because the redis they were connected to was >= 2.5) will get "downgraded" to the basic lock when they upgrade this plugin.

I would prefer an implementation based on `SET NX/EX` to exist _along side_ the existing lua-based resilient lock index, not to replace it.

lib/resque/scheduler/locking.rb
carsonreinke: > I'm all for simplifying calls where appropriate.
> 
> In this case though, any user who previously qualified for the resilient lock (because the redis they were connected to was >= 2.5) will get "downgraded" to the basic lock when they upgrade this plugin.
> 
> I would prefer an implementation based on `SET NX/EX` to exist _along side_ the existing lua-based resilient lock index, not to replace it.

Good point and good catch on the Redis version.  Still kind of working on this, as there is some spec failures too.

lib/resque/scheduler/lock/resilient_modern.rb
```diff
@@ -0,0 +1,14 @@
+# vim:fileencoding=utf-8
+require_relative 'base'
+
+module Resque
+  module Scheduler
+    module Lock
+      class ResilientModern < Resilient
```
carsonreinke, line 7: I couldn't come up with a better name

lib/resque/scheduler/locking.rb
```diff
@@ -97,7 +104,7 @@ def master_lock_key
       end
 
       def redis_master_version
-        Resque.data_store.redis.info['redis_version'].to_f
+        Gem::Version.new(Resque.data_store.redis.info['redis_version'])
```
iloveitaly, line 107: Why the change to using `Gem::Version` here?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ## [Unreleased]
 ### Changed
 - Remove support for Ruby < 2.3
+- Add `Lock::ResilientModern` that uses GET with NX/EX instead of Lua script to set master lock
```
iloveitaly, line 8: Can we link to this PR in the changelog?

lib/resque/scheduler/lock/resilient_modern.rb
iloveitaly: What does inheriting from `Resilient` give us?

I'd prefer inheriting from `base` and add a implementation for `locked?`. This would make it easier to remove the `Resilient` (and `Basic`) lock implementations in the future, which I can't imagine we wouldn't do now that we have this awesome new lock implementation (and redis `2.x` is very old at this point).

iloveitaly (changes requested):

iloveitaly:
Thanks for this great change! Excited to get this pulled in.

lib/resque/scheduler/locking.rb
carsonreinke: @iloveitaly this offers "native" version number parsing/comparison which was needed to see when the Redis feature was available.  Specifically, the patch level comparison.

CHANGELOG.md
carsonreinke: Sure thing

lib/resque/scheduler/lock/resilient_modern.rb
carsonreinke: @iloveitaly The `Resilient` implementation offers the `locked?` method which retrieves and renews the lock.  So, it looks like in Redis 6.2 a [similar command](https://redis.io/commands/getex) was added.

@yaauie suggested I retain the existing implementation to avoid users downgrading to a worse master lock implementation if they upgraded resque-scheduler.

One possibility could be switching the Redis requirement to 6.2 and completely removing the inheritance and Lua script implementation.

PR #735, opened 2021-11-21
Fix reading configuration from environment for boolean values

### LINKED ISSUE #565: Logging quiet when backgrounded even when logfile is set

carsonreinke:
Fixes #565 based on comments in #703.

Copied the truthy values from Rails (well modified them).  This retains the default of false unless proving truth.


lib/resque/scheduler/configuration.rb
```diff
@@ -8,56 +8,61 @@ def configure
         yield self
       end
 
+      attr_writer :environment
+      def environment
+        @environment ||= ENV
```
iloveitaly, line 13: Why not use `ENV` directly here? So folks can customize the configuration without modifying ENV?

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 ### Changed
 - Remove support for Ruby < 2.3
 
+### Fixed
+- Reading configuration from environment for boolean values
```
iloveitaly, line 10: Could we improve this description? The functionality change here is we aren't testing for existence anymore and have specific boolean-looking strings that we look for, correct?

iloveitaly (changes requested):

lib/resque/scheduler/configuration.rb
carsonreinke: @iloveitaly it just makes it easier to write tests to allow this be provided instead of mocking/stubbing.

CHANGELOG.md
carsonreinke: @iloveitaly updated!

carsonreinke:
:tada:

PR #736, opened 2021-11-21
Unblock rufus-scheduler upgrade by fixing shutdown

carsonreinke:
Updates #723

Looking into this more, it appears the use of calling `#join` on the scheduler is not needed when calling `#shutdown(:wait)` since that will block the current thread until shutdown.  The reason this worked before looks like simply `rufus-scheduler` unexpectedly supported this and it appears completely accidental.  Looking into the differences it looks the reason why is that the old `rufus-scheduler` method allowed calls on `#join` after shutdown because the `@thread` variable is never reset after shutdown (see https://github.com/jmettraux/rufus-scheduler/blob/v3.2.0/lib/rufus/scheduler.rb#L171).  After changes in v3.6.0, this is no longer allowed since it is using a queue system instead.

### DIFF
--- a/lib/resque/scheduler.rb
+++ b/lib/resque/scheduler.rb
@@ -377,7 +377,6 @@ module Resque
 
       def stop_rufus_scheduler
         rufus_scheduler.shutdown(:wait)
-        rufus_scheduler.join
       end
 
       def before_shutdown
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Remove support for Ruby < 2.3
 - Configuration to use a set of truthy values to enable boolean settings instead of simply existence
 - Add `delay_or_enqueue_at` for delaying existing jobs or creating a new job
+- Unblock rufus-scheduler lock on lower than 3.7 by fixing scheduler shutdown
 
 ## [4.5.0] - 2021-09-25
 ### Added
--- a/resque-scheduler.gemspec
+++ b/resque-scheduler.gemspec
@@ -61,5 +61,5 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'resque', '>= 1.27'
   # rufus-scheduler v3.7 causes a failure in test/multi_process_test.rb
   # rufus-scheduler v3.3 is missing a to_local method which fails tests
-  spec.add_runtime_dependency 'rufus-scheduler', '~> 3.2', '< 3.7', '!= 3.3'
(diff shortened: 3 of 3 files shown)


TomA-R:
Thanks @carsonreinke! This would be really good to get merged, thanks in advance 🙇 cc @iloveitaly

fagiani:
@iloveitaly I have confirmed those changes get tests passing and allowing dependencies to get upgraded. Is there anything else pending to get this PR merged? Keep Rocking!

iloveitaly:
@fagiani I've upgraded CI to do a matrix test across a bunch of rufus & resque versions. Can you rebase on master to fix merge conflicts and ensure this is tested across different configurations?

fagiani (changes requested):
This needs to be rebased to master in order to merge conflicts and pass on CI workflows. Would you mind allowing changes to the PR and allowing @fagiani as a contributor?

fagiani:
@iloveitaly I've tried to set myself as a reviewer in order to check if I could push the rebased branch but as far as I was able to find, in order to achieve that the PR must allow changes from maintainers and I need to become a contributor somehow. In respect to @carsonreinke's work I did not recreate a new PR. Please let me know how would you like to proceed.

carsonreinke:
@iloveitaly @fagiani I pushed up a rebase, let me know if you need anything else.

lib/resque/scheduler.rb
```diff
@@ -377,7 +377,6 @@ def handle_signals_with_operation
 
       def stop_rufus_scheduler
         rufus_scheduler.shutdown(:wait)
-        rufus_scheduler.join
```
fagiani, line 380: Looking at the CI tests, looks like the only version failing is `3.6`. 

I'd suggest something like:

```ruby
rufus_scheduler.join if Gem.loaded_specs["rufus_scheduler"].version.to_s.starts_with?("3.6")
```

I'd guess this change would potentially fix the test.

fagiani (changes requested):

lib/resque/scheduler.rb
carsonreinke: @fagiani ugh, looks like a flake, rebuilt and it passed.  Ran it locally without failure.  I don't believe that failure would even be related to this change though.

fagiani: great @carsonreinke ! 

@iloveitaly when is the next release including this PR planned?

Keep Rocking!

TomA-R:
Thanks for merging this, @iloveitaly! Any chance we could get this released? Thanks in advance

PR #752, opened 2022-09-21
Use non-deprecated form of Redis.sadd/srem

irvingreid:
This fixes the deprecation warnings for `Redis#sadd` and `Redis#srem` output by version 4.8 of `redis-rb`, following the example of the changes made to Resque in https://github.com/resque/resque/pull/1827

Other tests failed when upgrading `redis-rb` to 5.0.4, which is why I added the restriction to `resque-scheduler.gemspec`

To avoid future regressions, I enabled `Redis.raise_deprecations = true` so that the test suite is more likely to catch future Redis deprecations.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -29,6 +29,7 @@ Resque Scheduler authors
 - Henrik Nyh
 - Hormoz Kheradmand
 - Ian Davies
+- Irving Reid
 - James Le Cuirot
 - Jarkko Mönkkönen
 - Jimmy Chao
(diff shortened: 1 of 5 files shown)


irvingreid:
Looks like your test matrix pairs Redis 4.8 with older versions of the `resque` gem, which trigger deprecation failures. Would you rather change the test matrix, or remove `Redis.raise_deprecations = true`?

irvingreid:
@PatrickTulskie removed the gemspec change here too, because the main branch fixed Redis 5.0 compatibility :-)

carlwiedemann:
Friendly bump. Is there anything preventing this from getting merged?

PR #753, opened 2022-09-21
Fix `gethostbyaddr` deprecation warnings

irvingreid:
Use supported Addrinfo APIs to get the fully qualified local host name.

### DIFF
--- a/lib/resque/scheduler/lock/base.rb
+++ b/lib/resque/scheduler/lock/base.rb
@@ -47,7 +47,7 @@ module Resque
 
         def hostname
           local_hostname = Socket.gethostname
-          Socket.gethostbyname(local_hostname).first
+          Addrinfo.getaddrinfo(local_hostname, 'http').first.getnameinfo.first
         rescue
           local_hostname
         end


irvingreid:
After a fair bit of local debugging, I suspect this test failure is flaky? I could only reproduce it when I had another instance of `resque-scheduler` attached to my Redis server.

PatrickTulskie:
Hey I think this is probably okay to sneak into a patch release. Would you mind removing the gemspec changes? The gemspec changes aren't required for this fix.

irvingreid:
> Would you mind removing the gemspec changes? The gemspec changes aren't required for this fix.

@PatrickTulskie done, thanks for reviewing!

PatrickTulskie:
Hey it looks like those commits are still in this branch. Can you remove the out of scope commits instead of just reverting them? I can also do a merge but the rebase with a single clean commit is better. Gonna aim to get in a point release next week.

irvingreid:
Ah OK. We usually do squash merges here, so we don't fuss about the branch being noisy :-)

irvingreid:
Sorry about the delay - I got tricked by Github not syncing the `upstream` and `origin` master branches quite the way I expected, and couldn't figure out why the test was failing (and couldn't figure out how to re-run that one test case without pushing a change - is that possible with this CI setup?)

oehlschl:
@PatrickTulskie looks like this should be good to go now -- one commit as requested, and checks are passing. Can you please add this to a coming release for Ruby 3 compatibility?

PatrickTulskie (approved):

PatrickTulskie:
Sorry about the delay merging this. Was dealing with an end of quarter crunch. Anyhow, it's in gem v4.8.0 shipped just now.

PR #756, opened 2022-09-25
Use `Resque::DataStore#reconnect` for redis-rb v5.0

mishina2228:
This PR will fix a part of #750.

`Redis::Client#reconnect` has been removed from redis-rb v5.0.0.
This PR replaces it with `Resque::DataStore#reconnect`.
It is available since Resque v1.27.0, so it is backward compatible.

- Resque v1.27.0
  https://github.com/resque/resque/blob/v1.27.0/lib/resque/data_store.rb#L80-L83
  ```ruby
      # Force a reconnect to Redis.
      def reconnect
        @redis.client.reconnect
      end
  ```
- Resque v2.4.0
  https://github.com/resque/resque/blob/v2.4.0/lib/resque/data_store.rb#L79-L85
  ```ruby
      # Force a reconnect to Redis without closing the connection in the parent
      # process after a fork.
      def reconnect
        @redis.disconnect!
        @redis.ping
        nil
      end
   ```

PR #757, opened 2022-09-25
Fix tests for compatibility with redis-rb v5.0

mishina2228:
This PR fixes failing tests with redis-rb v5.0,
and also added matrix for redis-rb v4 and v5.

Here are the details:

- Field and value for hash must be String, Integer or Float
  - redis-rb 4.x used to call `to_s` on arguments, but redis-rb 5.x won't.
  - See here: https://github.com/redis/redis-rb/issues/1143#issuecomment-1231365756
- Use `Resque::DataStore#reconnect` instead of removed `Redis::Client#reconnect`
  - #756 already has this change, 
    but I would like to make all CI pass, so I included the same one.
- Added matrix redis version
  - Test with redis-rb v4.x and v5.x
  - Update resque test matrix from `~> 2.2.0` to `~> 2.4.0`
  - Exclude the combination of the following:
    - redis-rb v5.x with Ruby 2.4 or older
    - redis-rb v5.x with Resque 1.27.x
  - Fix resque-version in exclusions
    - They should be started with `~> `.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -28,30 +28,42 @@ jobs:
           - 3.1
         resque-version:
           - "master"
-          - "~> 2.2.0"
+          - "~> 2.4.0"
           - "~> 1.27"
         rufus-scheduler:
           - "3.2"
           - "3.4"
           - "3.5"
           - "3.6"
+        redis-version:
+          - "~> 4.x"
+          - "~> 5.x"
         exclude:
           - ruby-version: head
             rufus-scheduler: 3.2
 
           - ruby-version: 2.3
-            resque-version: 1.27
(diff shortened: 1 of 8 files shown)


mishina2228:
As stated at https://github.com/resque/resque-scheduler/pull/753#issuecomment-1254110430, some of the tests seem flaky.

mishina2228:
OK, I found the cause of the problem.
It seems that the tests fail if data from the previous test case is still in Redis.
I made sure that `flushall` is executed like other test files.

PatrickTulskie (approved):

PatrickTulskie:
Having some CI/branch protection issues. Lemme get that sorted out with @iloveitaly and then I can get this into master and a new release out.

PR #760, opened 2022-10-07
Update code sample on reconnect

mishina2228:
refs: #756
This PR updates the sample code in README.md to use `Resque::DataStore#reconnect` instead of `Redis::Client#reconnect`.


PatrickTulskie (approved):
Thanks!

PR #762, opened 2022-12-08
Replace `Socket.gethostbyname` with `Addrinfo.getaddrinfo`

kyoshidajp:
Since `Socket.gethostbyname` is already deprecated, replaced it with `Addrinfo.getaddrinfo`

Ref: https://bugs.ruby-lang.org/issues/13097


mishina2228:
Duplicate of #753

PR #765, opened 2023-01-21
Replace gethostbyname with Addrinfo.getaddrinfo

tylerdiaz:
This addresses the depreciation warning:

```
/gems/resque-scheduler-4.5.0/lib/resque/scheduler/lock/base.rb:50: warning: Socket.gethostbyname is deprecated; use Addrinfo.getaddrinfo instead.
```


PatrickTulskie:
FYI this was fixed in #753 and it's in v4.8.0 released today.

PR #766, opened 2023-03-01
Update CI matrix and fix tests

mishina2228:
This PR tweaks test codes.
- Add Ruby 3.2 to the CI matrix
  - I wanted to add ruby-head as well, but decided against it since mono_logger issue mentioned at https://github.com/resque/resque/issues/1856 has not yet been resolved.
- Ensure that Sinatra > 2.0 is used
  - For some reason, Sinatra 1.0 may be installed.
    If 1.0 is used, the following error occurs.
    ```
    <internal:/opt/hostedtoolcache/Ruby/3.2.1/x64/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:37:in `require': cannot load such file -- rack/showexceptions (LoadError)
    Did you mean?  rack/show_exceptions
    ```
- Copy `resque/server/test_helper` from Resque and use it
  - I removed this file at https://github.com/resque/resque/pull/1851 because it was not used in Resque side, but I found that it was used in this project.
  I believe the Resque changes should be left as is, as it is only needed for testing.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -26,6 +26,7 @@ jobs:
           - 2.7
           - "3.0"
           - 3.1
+          - 3.2
         resque-version:
           - "master"
           - "~> 2.4.0"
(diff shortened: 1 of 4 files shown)


iloveitaly:
Thanks for the contribution!

PR #768, opened 2023-04-13
Fixed Queue now button misalignment

ytkg:
Fixed Queue now button misalignment.

I referred to:
https://github.com/resque/resque-scheduler/blob/121e3427d1211baf8354067321d1f1dd14c7e4a1/lib/resque/scheduler/server/views/scheduler.erb#L42-L43

## before
image

## after
image


PatrickTulskie:
@ytkg I think this might be fixed in the latest release. Can you have a look when you get a second and update/close this PR? Thanks!

PR #769, opened 2023-05-31
Version 4.9.0

PatrickTulskie:
Updated changelog and version for 4.9.0


PatrickTulskie:
For anyone following along, I'm trying to deal with specs failing from what appears to be deprecation warnings. Looking into it before the release.

PR #781, opened 2023-12-04
Fix json dump crasher with JSON 2.7.0+

PatrickTulskie:
Attempting to fix a crasher that primarily affects Ruby 3+ when using JSON 2.7.0 when dumping logs to JSON.

Thanks @mishina2228 for pointing out this is related to https://github.com/flori/json/issues/553

### DIFF
--- a/lib/resque/scheduler/logger_builder.rb
+++ b/lib/resque/scheduler/logger_builder.rb
@@ -59,13 +59,14 @@ module Resque
       def json_formatter
         proc do |severity, datetime, progname, msg|
           require 'json'
-          JSON.dump(
+          log_data = {
             name: progname,
             progname: progname,
             level: severity,
             timestamp: datetime.iso8601,
             msg: msg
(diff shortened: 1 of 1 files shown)


PatrickTulskie:
I'll squash the commits before this gets merged. Just doing this work from the GitHub interface and my phone while I'm on the road.

mishina2228:
FYI, the reason for the crash is json 2.7.0 rather than Ruby 3+.
https://github.com/flori/json/issues/553

PatrickTulskie:
> FYI, the reason for the crash is json 2.7.0 rather than Ruby 3+. [flori/json#553](https://github.com/flori/json/issues/553)

Thank you for that. I was scratching my head trying to figure out why this wasn't a more widespread issue and doesn't affect any of the apps I'm running in prod. I don't use JSON logging with resque-scheduler so I figured that was the case. Lemme update the PR and commits with this info though.

Issue #700, opened 2020-07-06
How to configure Lock timeout


allensds:
Hi,

The base class of the Lock constructor suggests the timeout is configurable:

`@timeout = options[:timeout] || 60 * 3`

However, in the locking module, there is no way to put in the options with code like this. How do we easily change the configuration without changing the source code?

     def build_master_lock
        if supports_lua?
          Resque::Scheduler::Lock::Resilient.new(master_lock_key)
        else
          Resque::Scheduler::Lock::Basic.new(master_lock_key)
        end
      end

PR #786, opened 2023-12-19
Expose timeout for lock via environment variable configuration

pmm4654:
Expose the `timeout` configuration option used in `Resque::Scheduler::Locking::Base`

Related issue: https://github.com/resque/resque-scheduler/issues/700

### DIFF
--- a/lib/resque/scheduler/env.rb
+++ b/lib/resque/scheduler/env.rb
@@ -54,6 +54,7 @@ module Resque
         at_exit { cleanup_pid_file }
       end
 
+      # rubocop:disable Metrics/AbcSize
       def setup_scheduler_configuration
         Resque::Scheduler.configure do |c|
           c.app_name = options[:app_name] if options.key?(:app_name)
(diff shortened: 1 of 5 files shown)


learoyklinginsmith:
@PatrickTulskie any chance we could get some eyes on this change? 🙏

lib/resque/scheduler/env.rb
```diff
@@ -56,21 +56,21 @@ def setup_pid_file
...
         Resque::Scheduler.configure do |c|
-          c.app_name = options[:app_name] if options.key?(:app_name)
-
-          c.dynamic = !!options[:dynamic] if options.key?(:dynamic)
-
-          c.env = options[:env] if options.key?(:env)
-
-          c.logfile = options[:logfile] if options.key?(:logfile)
-
-          c.logformat = options[:logformat] if options.key?(:logformat)
```
PatrickTulskie, line 67: Can we just leave this and add the new timeout option?

PatrickTulskie (changes requested):
Overall good. I'd just like to set the env stuff the way it was instead of the meta programming approach. Calling `send` usually winds up causing a lot more scrutiny.

pmm4654:
@PatrickTulskie Hey there! I'm just nudging this PR after I made the changes you requested.

PatrickTulskie (approved):

PR #799, opened 2025-01-18
Require rack < 3

and9000:
As resque doesn't support rack >=3 and as it doesn't specify its required versions correctly, this PR is needed to make resque-scheduler's CI green


PatrickTulskie (approved):
LGTM. Should be fine as a development dependency right now. There's some work being done on fixing this for Resque so I'll lift the restriction here when that's done.

PR #812, opened 2025-12-25
Fix double invocation of schedule hooks for delayed jobs

### LINKED ISSUE #810: after_schedule hook is called during handle_delayed_items re-enqueue
## Description 

When resque-scheduler processes delayed jobs via `handle_delayed_items` (moving them from the delayed queue to the regular Resque queue), it calls both `before_delayed_enqueue` **_and_** `after_schedule` hooks. This is unexpected because `after_schedule` should only be called when a job is initially scheduled, not when it's being moved to the regular queue. That's also what the Readme states. 

##  Expected Behavior

This is what I would say is expected:
 
- `after_schedule` should be called once when a job is scheduled via `Resque.enqueue_at`
- `before_delayed_enqueue` should be called once when the job is moved from the delayed queue to the regular queue
- These hooks should be mutually exclusive - a job is either being scheduled _or_ being enqueued

It could of course also be that I am misunderstanding or missing something.

##  Actual Behavior

When `handle_delayed_items` processes a delayed job:
 
1. `before_delayed_enqueue` is called
1. `after_schedule` is also called (incorrect - the job is not being scheduled, it's being enqueued)

##  Root Cause

The issue as I see it is in `enqueue_from_config` (`scheduler.rb:334`) which wraps the enqueue operation in `process_schedule_hooks`:

```
Resque::Scheduler::Plugin.process_schedule_hooks(klass, *params) do
  Resque::Scheduler::Plugin.run_before_delayed_enqueue_hooks(klass, *params)
  # ... enqueue job
end
```

and `process_schedule_hooks` (`plugin.rb:6-13`) unconditionally calls `run_after_schedule_hooks`:
```
  def self.process_schedule_hooks(klass, *args)
    return unless run_before_schedule_hooks(klass, *args)
    yield
    run_after_schedule_hooks(klass, *args)
  end
```

##  Reproduction Script

```
#!/usr/bin/env ruby
#
# Standalone script to demonstrate a bug in resque-scheduler where the
# `after_schedule` hook is called during `handle_delayed_items`.
#
# Prerequisites:
#   Redis running on localhost:6379
#
# Run with: ruby <filename.rb>

require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "resque", "~> 2.6"
  gem "resque-scheduler", "~> 4.11"
end

require "resque"
require "resque-scheduler"

redis = Redis.new(db: 15)
Resque.redis = redis
redis.flushdb

# Test job that tracks hook calls
class TestHookJob
  @queue = :default
  @hook_counts = {after_schedule: 0, before_delayed_enqueue: 0}

  class << self
    attr_reader :hook_counts
  end

  def self.perform(*args)
    puts "Job performed with args: #{args.inspect}"
  end

  def self.after_schedule(*args)
    @hook_counts[:after_schedule] += 1
    puts "HOOK: after_schedule called (count: #{@hook_counts[:after_schedule]})"
  end

  def self.before_delayed_enqueue(*args)
    @hook_counts[:before_delayed_enqueue] += 1
    puts "HOOK: before_delayed_enqueue called (count: #{@hook_counts[:before_delayed_enqueue]})"
  end
end

puts "=== Step 1: Schedule a job 1 hour from now ==="
Resque.enqueue_at(Time.now + 3600, TestHookJob, "test_arg")
puts ""

puts "=== Step 2: Simulate time passing and process delayed items ==="
# ########
# Expected: only before_delayed_enqueue should run here.
# Actual: after_schedule is being called again (bug).
# ########
Resque::Scheduler.handle_delayed_items(Time.now + 4000)
puts ""

puts "=== Summary ==="
puts "Hook call counts:"
puts "  after_schedule:         #{TestHookJob.hook_counts[:after_schedule]} (expected: 1)"
puts "  before_delayed_enqueue: #{TestHookJob.hook_counts[:before_delayed_enqueue]} (expected: 1)"

# Cleanup
redis.flushdb
```

I ran this with:

- resque-scheduler version: 4.11.0
- resque version: 2.7.0
- ruby version: 3.4.4

##   Suggested Fix
I would say `enqueue_from_config` should not call `process_schedule_hooks` since the job is being enqueued, not scheduled. It should only call `run_before_delayed_enqueue_hooks` without the wrapping `process_schedule_hooks`.

As probably many projects already rely on the hooks being this way, there could as well be a configuration flag
```
Resque::Scheduler.configure do |config|
  config.skip_after_schedule_on_enqueue = true  # defaults to false
end
```

Also maybe a separate `after_delayed_enqueue` hook could be the solution, additionally to the current `after_schedule` and `before_delayed_enqueue`.

Let me know what you think. Maybe this is intended and/or I might misunderstand or miss something.

top-sigrid:
Fixes [#810](https://github.com/resque/resque-scheduler/issues/810)

  ## Summary
  - Move invoking schedule hooks from `enqueue_from_config` to `enqueue_recurring`
  - This ensures hooks are called for cron jobs but not again when delayed jobs are transferred to the work queue
  - Extract `extract_klass_and_params` helper to avoid duplication
  - Update README to clarify when hooks are called

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -39,7 +39,7 @@ Metrics/MethodLength:
 # Offense count: 2
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 364
+  Max: 375
 
 # Offense count: 1
 Style/CaseEquality:
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -56,6 +56,7 @@ Resque Scheduler authors
 - Nickolas Means
 - Olek Janiszewski
 - Olivier Brisse
+- Paul Wille
 - Petteri Räty
 - Phil Cohen
 - Rob Olson
--- a/README.md
+++ b/README.md
@@ -502,12 +502,13 @@ A future version of resque-scheduler may do this for you.
 Similar to the `before_enqueue`- and `after_enqueue`-hooks provided in Resque
 (>= 1.19.1), your jobs can specify one or more of the following hooks:
 
-* `before_schedule`: Called with the job args before a job is placed on
-  the delayed queue. If the hook returns `false`, the job will not be placed on
-  the queue.
-* `after_schedule`: Called with the job args after a job is placed on the
-  delayed queue. Any exception raised propagates up to the code with queued the
-  job.
+* `before_schedule`: Called with the job args before a job is scheduled. This
+  includes delayed jobs (when placed on the delayed queue via `enqueue_at`/`enqueue_in`)
+  and recurring cron jobs (each time the job is enqueued).
+  If the hook returns `false`, the job will not be enqueued.
+* `after_schedule`: Called with the job args after a job is scheduled (same
+  scenarios as `before_schedule`). Any exception raised propagates up to the
+  calling code.
 * `before_delayed_enqueue`: Called with the job args after the job has been
   removed from the delayed queue, but not yet put on a normal queue. It is
   called before `before_enqueue`-hooks, and on the same job instance as the
--- a/lib/resque/scheduler.rb
+++ b/lib/resque/scheduler.rb
@@ -296,16 +296,7 @@ module Resque
 
       # Enqueues a job based on a config hash
       def enqueue_from_config(job_config)
-        args = job_config['args'] || job_config[:args]
-
-        klass_name = job_config['class'] || job_config[:class]
-        begin
-          klass = Resque::Scheduler::Util.constantize(klass_name)
-        rescue NameError
-          klass = klass_name
-        end
-
-        params = args.is_a?(Hash) ? [args] : Array(args)
+        klass, klass_name, params = extract_klass_and_params(job_config)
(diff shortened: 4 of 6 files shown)


top-sigrid:
Hi @PatrickTulskie, just a friendly ping on my open PRs (#812, #813, #817). They've been sitting for a while now - is there anything I can do to help move them along? Happy to rebase onto the 5.0.0 release if that helps.

I also just opened #822 (pipelining `count_all_scheduled_jobs`) which fixed a real production bottleneck for us. Would be great to get that one in as well.

top-sigrid:
Hi @PatrickTulskie, I also rebased this PR. But maybe have in mind what i proposed in https://github.com/resque/resque-scheduler/issues/815. I still think, having dedicated `before/after_recurring` hooks would actually be the cleaner solution.

PatrickTulskie:
@top-sigrid commenting here too to keep things linked up... I think we should do what you suggested and add different hooks for the recurring as described here https://github.com/resque/resque-scheduler/issues/815

PR #813, opened 2025-12-25
Fix flaky test by using strings instead of symbols for JSON-serialized args

top-sigrid:
Fixes a flaky test on Ruby 2.4 where mocha expectations fail due to symbol/string mismatch.


top-sigrid:
@PatrickTulskie I also rebased this, with ruby versions `2.*` not being supported anymore, these aren't flaky anymore, so I leave this up to you, if this still makes sense :)

PatrickTulskie (approved):

PR #817, opened 2026-01-11
Fix schedule hooks firing when timestamp <= now

top-sigrid:
When `enqueue_at` is called with a timestamp in the past or equal to now, the job bypasses the delayed queue and goes directly to the work queue. Previously, `before_schedule` and `after_schedule` hooks were incorrectly called in this case.

This fix moves `process_schedule_hooks` inside the `else` branch so hooks only fire when jobs actually enter the delayed queue.

Fixes #816

### DIFF
--- a/lib/resque/scheduler/delaying_extensions.rb
+++ b/lib/resque/scheduler/delaying_extensions.rb
@@ -21,19 +21,24 @@ module Resque
       # timestamp has passed. It respects Resque.inline option, by
       # creating the job right away instead of adding to the queue.
       def enqueue_at_with_queue(queue, timestamp, klass, *args)
-        plugin.process_schedule_hooks(klass, *args) do
-          if Resque.inline? || timestamp.to_i <= Time.now.to_i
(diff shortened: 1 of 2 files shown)


PatrickTulskie:
@top-sigrid I think this is fine to merge, but it could use a rebase. Mind rebasing on master, pushing it up and let's see if CI passes.

top-sigrid:
Hi Patrick, thanks for getting back to this! I can definitely do that, will find the time very soon. Probably a good idea to do this for my other PRs aswell :)

PatrickTulskie:
@top-sigrid oh yeah actually you're right there are a few that we can probably merge in once they are rebased. I'm aiming to get a few resque releases out around mid-month. If we can get them in before that they'll make the release. Holler when you want me to check them!

top-sigrid:
Hi @PatrickTulskie I finally found the time to dedicate some time to this again. This one is rebased and the CI passes so should be fine. Let me know if you have any further wishes or review :)

